PR #40, opened 2019-01-07
Updated Tutorial and supporting docs

JasonStoltz:
- Updated Tutorial and supporting documentation to be more open ended. This is to encourage the uses of READMEs to figure out how to accomplish tasks.
- Added documentation for React Components.
- Fixed a bug in the README
- Changed signature of Facet.js so that default values is more apparent


packages/react-search-ui/README.md
```diff
@@ -69,10 +69,218 @@ Connectors are explained in the top level [README](../../README.md). The first
...
+- [PagingInfo](#c_paginginfo)
+- [Results](#results)
+- [ResultsPerPage](#c_resultsperpage)
+- [SearchBox](#c_searchbox)
+- [Sorting](#c_sorting)
+
+#### ErrorBoundary<a id="c_errorboundary"></a>
+
+Use this to handle unexpected errors. Any content passed to this component will
+be show unless an unexpected Error is thrown. it will then replaced with an
```
brianearwood: change 'show' to 'shown'

packages/react-search-ui/README.md
```diff
@@ -69,10 +69,218 @@ Connectors are explained in the top level [README](../../README.md). The first
...
+<ResultsPerPage />
+```
+
+#### SearchBox<a id="c_searchbox"></a>
+
+Properties:
+
+| Name       | type      | Required? | Default                                                | Options | Description                                           |
+| ---------- | --------- | --------- | ------------------------------------------------------ | ------- | ----------------------------------------------------- |
+| inputProps | Object    | no        |                                                        |         | Props of 'input' element to be passed through.        |
```
brianearwood: I'm confused about what the options are in this case

JasonStoltz: @brianearwood Updated, does that make more sense?

brianearwood: Is `placeholder` the only option?

JasonStoltz: It could be any attribute of an input element

Issue #44, opened 2019-01-17
New URL filter scheme


JasonStoltz:
How do we represent search parameters from the App Search API in a query string? 

Challenge: need to flatten highly nested structure and maintain typing.

Reference: https://swiftype.com/documentation/app-search/api/search/filters

## API Schema

```
{
  “all”: [{
     // Range with type Number
     {"visitors": {
        "from": 1,
        "to": 100
     }},
     // Range with type Date
     {"date_established": {
       "from": "1900-01-01T12:00:00+00:00",
       "to": "1950-01-01T00:00:00+00:00"
     }},
     // Geo with type Geolocation
     {"location": {
       "center": "37.386483, -122.083842",
       "distance": 300,
       "unit": "km"
     }},
     // Value with type Text
     {"world_heritage_site": "true"},
     // Value with type Date
     {“created_date”: "1900-01-01T12:00:00+00:00"},
     // Value with Type Number
     {“visitors”: 100},
     // Combining with "all"
     {“category”: ”fun”},
     {“category”: ”healthy”}
     // Nested "or"s
     {“category”: [”gradeA”, “gradeB”]} 
  }],
  // Combining with "any"
  “or”: { … },
  // Combining with "none"
   “none”: { … },
}
```

## Proposed Query String Schema

f_{filterType}{valueType}{andAnyNone}{_sequence}_{fieldName}{[]}={value}

- filterType: 
  - “v” for Value
  - “r” for Range
  - “g” for Geo.
- valueType:
  - “t” for Text”
  - “d” for Date
  - “n” for Number
  - “g” for Geolocation
- andAnyNone: 
  - “a” for “and”
  - “o” for “any”
  - “n” for “none”
- _sequence: Nothing present if there is only one value, “_1” for the second, then “_2”, etc.
- fieldName
- value:
  - Value filters - just the value
  - Range filter - {low}_{high}
  - Geo filter - {center}_{distance}_{unit}

## Examples

Range with type Number

```
{"visitors": {
   "from": 1,
  "to": 100
}}
     
f_rna_visitors=1_100
```

Range with type Date

```
{"date_established": {
  "from": "1900-01-01T12:00:00+00:00",
  "to": "1950-01-01T00:00:00+00:00"
}}

f_rda_date_established=1900-01-01T12:00:00+00:00_1950-01-01T00:00:00+00:00
```

Geo with type Geolocation

```
{"location": {
 "center": "37.386483, -122.083842",
  "distance": 300,
[...]

PR #45, opened 2019-01-23
Reworked how query filters are maintained in the URL

### LINKED ISSUE #42: Numeric filter fails from query parameters
I have a dataset which includes a numeric field `difficulty`.

I want to filter on this, so I add a filter with `addFilter("difficulty", 2)`. This works great and filters my result set with the difficulty of 2. It also adds `?fv-difficulty=2` to my query parameters. If I then refresh my page I get the following error: 

```
Error: [400] Filters contains invalid value for field: difficulty; must be a number, an array of numbers, or a range object with `to` and/or `from` keys
    at handleErrorResponse (swiftype_app_search.umd.js:362)
```

JasonStoltz:
Fixes #42 #44 

The URL scheme I had for filters was insufficient. It needs to be able to handle the following scenarios:

```
{
  “all”: [{
     // Range with type Number
     {"visitors": {
        "from": 1,
        "to": 100
     }},
     // Range with type Date
     {"date_established": {
       "from": "1900-01-01T12:00:00+00:00",
       "to": "1950-01-01T00:00:00+00:00"
     }},
     // Geo with type Geolocation
     {"location": {
       "center": "37.386483, -122.083842",
       "distance": 300,
       "unit": "km"
     }},
     // Value with type Text
     {"world_heritage_site": "true"},
     // Value with type Date
     {“created_date”: "1900-01-01T12:00:00+00:00"},
     // Value with Type Number
     {“visitors”: 100},
     // Combining with "all"
     {“category”: ”fun”},
     {“category”: ”healthy”}
     // Nested "or"s
     {“category”: [”gradeA”, “gradeB”]} 
  }],
  // Combining with "any"
  “or”: { … },
  // Combining with "none"
   “none”: { … },
}
```

So, the query string need to be aware of the value type of filters, the Filter type, and OR vs AND vs NONE filters. Rather than coming up with some complicated schema for this in the query string, I chose to simply serialize the entire object using the [qs](https://github.com/ljharb/qs) library.

The only thing the qs library DOESN'T support is maintaining type. Meaning, if you serialize the Integer 1 to the query string, when deserialized it is serialized to the String "1".

To work around that, I added a custom encoder and decoder to simply maintain type by padding Number and Boolean values with "n_{value}_n" and "b_{value}_b" respectively.

For example, if I were to convert the object above using this new serializer, it would look like this:

```
filters[all][0][visitors][from]=n_1_n&filters[all][0][visitors][to]=n_100_n&filters[all][1][date_established][from]=1900-01-01T12%3A00%3A00%2B00%3A00&filters[all][1][date_established][to]=1950-01-01T00%3A00%3A00%2B00%3A00&filters[all][2][location][center]=37.386483%2C -122.083842&filters[all][2][location][distance]=n_300_n&filters[all][2][location][unit]=km&filters[all][3][world_heritage_site]=true&filters[all][4][created_date]=1900-01-01T12%3A00%3A00%2B00%3A00&filters[all][5][visitors]=n_100_n&filters[all][6][category]=fun&filters[all][7][category]=healthy&filters[all][8][category][0]=gradeA&filters[all][8][category][1]=gradeB
```

Note that while this DOES fix the issue with numeric value filters mentioned in #42, this does not fix existing issues in Search UI with Geo, Date Range, OR, and NONE filters. However, it does make the query string flexible enough to handle any of those use cases in the future.

latest

### DIFF
--- a/packages/search-ui/src/preserveTypesEncoder.js
+++ b/packages/search-ui/src/preserveTypesEncoder.js
@@ -1,0 +1,40 @@
+function isTypeNumber(value) {
+  return value && typeof value === "number";
+}
+
+function isTypeBoolean(value) {
+  return value && typeof value === "boolean";
+}
+
+function toBoolean(value) {
+  if (value === "true") return true;
+  if (value === "false") return false;
+  throw "Invalid type parsed as Boolean value";
+}
+
+/* Encoder for qs library which preserve number types on the URL. Numbers
+are padded with "n_{number}_n", and booleans with "b_{boolean}_b"*/
+
+export default {
+  encode(value, encode) {
+    if (isTypeNumber(value)) {
+      return `n_${value}_n`;
+    }
+    if (isTypeBoolean(value)) {
+      return `b_${value}_b`;
+    }
+    return encode(value);
+  },
+  decode(value, decode) {
+    //eslint-disable-next-line
+    if (/n_[\d\.]*_n/.test(value)) {
+      const numericValueString = value.substring(2, value.length - 2);
+      return Number(numericValueString);
+    }
+    if (/^b_(true|false)*_b$/.test(value)) {
+      const booleanValueString = value.substring(2, value.length - 2);
+      return toBoolean(booleanValueString);
+    }
+    return decode(value);
+  }
+};
--- a/packages/react-search-ui-views/src/types/ValueFilterValue.js
+++ b/packages/react-search-ui-views/src/types/ValueFilterValue.js
@@ -1,3 +1,3 @@
 import PropTypes from "prop-types";
 
-export default PropTypes.string;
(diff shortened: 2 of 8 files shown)


packages/search-ui/src/preserveTypesEncoder.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function toBoolean(value) {
+  if (value === "true") return true;
+  if (value === "false") return false;
+  throw "Invalid type parsed as Boolean value";
+}
+
+/* Encoder for qs library which preserve number types on the URL. Numbers
+are padded with "n_{number}_n", and booleans with "b_{boolean}_b"*/
```
byronhulcher: v cool

PR #79, opened 2019-02-06
Live search

JasonStoltz:
# Live Search

Finishes #79 

latest

## react-search-ui

```
<SearchBox
  searchAsYouType={true}
  debounceLength={500}
 />
```

- Added a `searchAsYouType` option to `SearchBox` that will perform requests as input is entered
- Added a `debounce` option to `SearchBox` to let a user balance the number of requests made to the server vs. how responsive the SearchBox is.
- Moved the `SearchBox` input value to be managed globally in the `searchTerm` state field, which lets the `SearchBox` be cleared from outside (like via a `reset` action), and also sets us up for query suggestions.

## search-ui

```
setSearchTerm("a term", {
  refresh: true,
  debounce: 500
})
```

- Added a `refresh` and `debounce` option to `setSearchTerm` action which correspond to the `searchAsYouType` and `debounce` options on the `SearchBox`.
- Debouncing URL state changes across the board to avoid getting an over abundance of URL state changes in browser history.
- Added "sequencing" to API calls so that old API calls are ignored if responses come back from the server out of order. This is especially a problem for the "Live Search" feature.
- Refactored tests.
- Updated `setSearchTerm` so that multiple calls to `setSearchTerm` can be made in quick succession. If a previous operation has actions locked, `setSearchTerm` will disregard that and make the request anyway. Other actions will respect an "isLoading" flag that disallows another action to be made before the previous action has completed.


packages/react-search-ui/README.md
```diff
@@ -600,18 +602,18 @@ ex.
...
+| `setResultsPerPage` | Integer                                                                                                                                                                                    |        |                                                                            |
+| `setSearchTerm`     | `searchTerm` String<br/> `options` Object<br/>`options.refresh` Boolean - Refresh search results on update. Default: `true`.<br/>`debounce` Number - Length to debounce if using `refresh` |        |                                                                            |
+| `setSort`           | `sortField` String - field to sort on<br/>`sortDirection` String - "asc" or "desc"                                                                                                         |        |                                                                            |
+| `trackClickThrough` | `documentId` String - The document ID associated with the result that was clicked<br/>`tag` - Array[String] Optional tags which can be used to categorize this click event                 |        | Report a clickthrough event. which is when a user clicks on a result link. |
```
scottybollinger: Last part of this sentence is weird. Not sure that period belongs there. "...event. which..."

packages/react-search-ui/README.md
```diff
@@ -600,18 +602,18 @@ ex.
...
-| `trackClickThrough` | `documentId` String - The document ID associated with the result that was clicked<br/>`tag` - Array[String] Optional tags which can be used to categorize this click event |        | Report a clickthrough event. which is when a user clicks on a result link. |
+| method              | params                                                                                                                                                                                     | return | description                                                                |
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------ | -------------------------------------------------------------------------- |
+| `addFilter`         | `name` String - field name to filter on<br/>`value` String - field value to filter on                                                                                                      |        | Add a filter in addition to current filters values                         |
```
scottybollinger: Any reason not to end all of these sentences in the descriptions with periods?

packages/search-ui/src/RequestSequencer.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  isOldRequest(request) {
+    return request < this.lastCompleted;
+  }
+
+  completed(request) {
+    this.lastCompleted = request;
+  }
+}
```
scottybollinger: This is nice!

JasonStoltz: 😎

packages/react-search-ui/README.md
JasonStoltz: Done.

JasonStoltz:
Good points @scottybollinger. Mind re-reviewing?

PR #86, opened 2019-02-12
Sandbox

JasonStoltz:
In addition to code review, please pull this repository down and get the sandbox application up and running to verify that this setup works.

Primary objectives:
- Add a "sandbox" application for local development purposes

Related changes I ended up making:
- Git ignoring `storybook-static` which is sometimes generated by Storybook and not meant to be in the git repo.
- Added a description of how Lerna is used to create this monorepo to give developers an idea of how to work with it.
- Split Lerna packages into `examples` and `packages`. `packages` being actual packages that are published to npm as part of `search-ui`, and examples just contains usage examples.
- Updated all commands to only act on actual  search-ui npm packages, and not example applications (when appropriate). For example, any command related to publishing should not actually attempt to publish an example app, only actual npm packages..
- Added a top-level `watch-js` command to watch and rebuild js files in all packages.
- Updated instructions for publishing. I was previously rewriting the version commit that was generated by `lerna` version, which broke the `lerna changed` command.
- Add the `sandbox` application along with instructions for running in the README. It is just a basic create-react-app generated react application with search-ui installed and set up.
- Moved common `devDependencies` out of the top level package.json file and into the `packages` directory, so that the dependencies won't be shared with example applications, which causes webpack build issues.
- Added configuration to `lerna.json` so that release commits always have the same commit message, and that markdown and tests are not considered "changed", which causes new versions of packages to be published even if they haven't changed.
- Changed lerna.json so that ALL packages are versioned together, and NOT independently.
- Updated all examples and the sandbox application to use a new key and engine, which is meant to be solely dedicated to running these examples.


brianearwood:
I managed to get it up and running after a few failed starts, which I think could be smoothed over by some extra descriptive text in the README file.

I left some notes as I was working:

1. Do I have to actually cd packages and then npm install or is it included in the original npm install?
2. My first run of `npm run bootstrap` gave me a few `Error: Cannot find module '@babel/core’` errors
3. After those errors, I went back and ran `npm install` inside of the packages directory and then `npm run bootstrap` worked

**In the sandbox README.md**
1. Not sure what directory I’m supposed to run `npm install bootstrap`
2. Also not sure about `npm run watch-js`
3. After running both of those commands in the overall `search-ui` directory, running `npm start` errors out
4. Running a global search for `start` in the repo surfaced a package.json file inside of `examples/sandbox` which worked.

**It works!**

README.md
```diff
@@ -36,6 +36,38 @@ _Note: The Search UI is in beta. We do not recommend production use._
...
+as "packages" in `lerna.json` so that `npx lerna bootstrap` will automatically wire up the
+examples to local dependencies.
+
+Because all examples are declared as "private", when running lerna commands other than bootstrap, (like `publish` and
+`test`), the `--no-private` flag should be appended.
+
+Dependencies are declared in a package.json hierarchy.
+
+- / package.json - Dependencies for repo tooling, like `husky` and `lerna`.
+- /packages package.json - Common dev dependencies for all Search UI npm packages. Any dev depenency that does not need to be called directly in a package level npm command
```
brianearwood: Misspelled `dependency`

README.md
```diff
@@ -36,6 +36,38 @@ _Note: The Search UI is in beta. We do not recommend production use._
...
+- /packages/{package_name} package.json - Package specific dependencies.
+
+Note that we do not encourage "hoisting" dependencies through lerna. This WILL
+cause the examples applications to error out from dependency version conflicts.
+
+### Installing dependencies
+
+```shell
+npm install
+(cd packages && npm install)
```
brianearwood: Not sure what the parenthesis means here. Should I run these commands separately?

After getting an error, I ran `npm install` inside of packages and it worked.

README.md
```diff
@@ -36,6 +36,38 @@ _Note: The Search UI is in beta. We do not recommend production use._
...
+npm run bootstrap
```
brianearwood: Not sure where to run this. I think throughout the guide, there is some confusion as to where I should be when I run commands.

examples/sandbox/README.md
```diff
@@ -0,0 +1,29 @@
...
+It is set up to use a pre-configured Elastic App Search engine.
+
+## Using
+
+To run:
+
+```
+# Bootstrapping with lerna will install all dependencies for this entire
+# repository.
+(cd ../.. && npm run bootstrap)
```
brianearwood: There could be more description in this section as to where I should start and where I should be when I run these commands.

If I'm in the `search-ui` folder, running the `cd ../..` puts me out in one of my computers containing folders which is not a git repo. Not sure how much of a problem this is for more experienced developers though I could see it being a hangup for less experienced ones.

examples/sandbox/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Start watching search-ui js for changes in the various library repositories
+(cd ../.. && npm run watch-js)
+
+# Run the web application
+npm start
```
brianearwood: I couldn't find any reference to `npm start` when looking through all of the `package.json` files until I ran a global search for "start" and realized that there is a new `examples` directory with the sandbox inside.

This is just related to the need for extra description around the structure of the mono-repo and how it all plays together.

examples/sandbox/README.md
```diff
@@ -0,0 +1,29 @@
+# Sandbox
+
+This example is a sandbox for local development. It sets up a basic page
+using Search UI.
+
+It is wired up to use the local source of the search-ui libraries as dependencies via lerna.
+This means as you update library code you can see it updated live in this project.
+
+It is set up to use a pre-configured Elastic App Search engine.
```
brianearwood: It would be nice to have some information about the engine and the dataset.

Considering the engine config: Since this is just a testbed for the Search UI, I'm not sure it's a place for people to change the engine config, what do you think?

README.md
JasonStoltz: It's meant to be a single shell command, literally run as `(cd packages && npm install)`.

JasonStoltz: Meaning, it's meant to be copied and pasted like that directly into your terminal and run

brianearwood: Ah, ok. I haven't seen the parentheses used a lot before.

README.md
JasonStoltz: Yeah, that was not clear at all.

examples/sandbox/README.md
JasonStoltz: @brianearwood I provided a bit more information about what the data set is.

> I'm not sure it's a place for people to change the engine config, what do you think?

I'm not quite sure what you're asking. But if you're developing locally, I think it would be pretty reasonable to change the engine config in whatever way you'd need. You might point it to an entirely different engine, or you might change the parameters... you just wouldn't be able to commit those changes back to the repository.

Did does that makes sense?

brianearwood: That makes sense.

I was wondering if we should point out with additional copy how people could easily swap out the default test engine credentials with their own to make the sandbox more meaningful for their use-case.

examples/sandbox/README.md
```diff
@@ -0,0 +1,48 @@
...
+This means as you update library code you can see it updated live in this project.
+
+It is set up to use a pre-configured Elastic App Search Engine. The data
+set is the "Sample Engine" data set provided as an example for App Search
+accounts. More info on that can be found in this [article](https://www.elastic.co/blog/a-walk-in-the-park-with-elastic-app-search-sample-engines).
+
+## Using
+
+First off, you'll need to get the dependencies installed correctly for this
+entire repository. To do this, there's a few commands you'll need to run that are listed
```
brianearwood: `there's` I think should be `there are`

examples/sandbox/README.md
JasonStoltz: Ah, great point. And actually, I think I could probably move the credentials to a .env file, so that you can update the engine and not have to worry about accidentally commiting your changes 👍

examples/sandbox/README.md
JasonStoltz: English is hard.

examples/sandbox/README.md
JasonStoltz: Just updated the instructions to include details about configuring a .env file.

And now that I'm thinking about this, once Reference UI is converted to use Search UI, I should totally just make the Sandbox an instance of the Reference UI.

That way you can configure much more than just the engine.

JasonStoltz:
@brianearwood Ready for another look.

brianearwood:
Very nice. Looks good.

Issue #90, opened 2019-02-14
Range facet filters won't show value if they don't have a name


JasonStoltz:
They will just show blank. Need to fallback to something else when a name is not provided.

JasonStoltz:
I am making the name "Required", so this will be a non issue.

Issue #15, opened 2018-11-14
Normalize the Data Model


JasonStoltz:
Come up with domain specific state shape that works best for the app, and not necessarily derived from API syntax.

The current syntax is derived from the App Search API syntax. It is highly nested and difficult to work with, especially when try to add or remove filters. This is because the API syntax is geared towards providing a DSL for querying, it was not designing with the intention of state management.

For this reason, a new state shape for managing UI state is would be better.

This will also push the connectors in a slightly different direction. Rather than converting to/from the App Search API syntax, they'll be converting to / from Search UI state syntax.

## Design

Note: Not modeling value types
    - Date, Text, Geo, Number

The App Search API does not distinguish between the Geo, Date, Text, and Number types. So likewise, the UI does not distinguish, because there is no way to tell what the schema is by looking at values in an API response. We just pass values through and let the API figure out what is what. This means that no code in Search UI can automatically detect a value as being a Date, for instance and format it or treat it as a date explicitly. Components will need to have optional flags that tell a component that they're dealing with Dates if they need to render values as dates, etc.

This is an overall positive approach I believe, because it's entirely possible that different backends will have different value types. For instance, we don't support a Boolean type, but other APIs may. This way, we just pass the value through and don't worry about the particular type, so it should "just work"

**Filter**
   - field: String - Which field is this filter associated with
   - type: **FilterType** - "all", "or", "none"
    - values: [**FilterValue**] - An array represents "or" filters

**Facet**
    - field: String - Which field is this facet associated with
    - type: String - "range" or "value"
    - data:** [**FacetValue**]

**FacetValue**
    - count: Number
    - value: **FilterValue** 

**FilterType: String["all" | "none" | "or"]**

** FieldValue: <String, Boolean, Number> **

**FilterValue > FilterValueRange\<T:FieldValue\>**
    - to: \<T\>
    - from: \<T\>
    - name: String - We'll need this to correlate applied dynamic Date Range filters with options in corresponding range Facets on a page reload. This will not be an optional field.

**FilterValue > FilterValueValue: FieldValue**

Cases to consider:
- Detecting if any filter value is applied to a particular field
- Not applying the same filter twice / uniqueness
- Switching between "all" and "or" for a field
- Deleting all filters values for a field
- Deleting an individual filter field
- Adding a new filter
- Creating specific types of filters via components, like a RangeFilter component with "from" and "to" fields.
- Facet filters

PR #103, opened 2019-02-21
Rework facets

JasonStoltz:
Finishes #90 
Finishes #8 
Finishes #15 

Facet filter components were updated to work with ANY facet values, regardless of type. Part of this change was updating how we model Facets and Filter Values: https://github.com/elastic/search-ui/issues/15. Since this affects the data model used all throughout this project, the changes were fairly extensive.

## To Do:
- [x] views / SingleSelectFacet needs to work with new types
- [x] views / MultiCheckboxFacet needs to work with new types
- [x] views / SingleLinksFacets needs to work with new types
- [x] Facet needs to work with new types
- [x] Driver needs to work with new types
- [x] Helper to derive a label from a FilterValue


examples/sandbox/src/App.js
```diff
@@ -48,6 +48,7 @@ export default function App() {
             }
           }
         },
+        disjunctiveFacets: ["acres"],
```
JasonStoltz: This was a small update to fix the sandbox application. The SingleSelectFacet needs a disjunctive facet to work properly.

packages/react-search-ui-views/package.json
```diff
@@ -69,6 +69,7 @@
   },
   "dependencies": {
     "autoprefixer": "^9.4.7",
+    "deep-equal": "^1.0.1",
```
JasonStoltz: I added this for Facet filters ... we needed this to compare an applied filter to determine which value from a facet the current.

I.e., if the filter `{from: 1, to: 100}` is applied to the `visitors` field, in the list of ranges in the `visitors` facet, we need to compare the values of the visitors facets to determine which is applied, by doing a deep equal. 

Since facet values could range fitlers, geo filters, or values, the deep equal will be able to compare regardless of the type of filter / facet.

packages/react-search-ui-views/src/SingleSelectFacet.js
```diff
@@ -19,27 +21,14 @@ Option.propTypes = {
   data: PropTypes.object.isRequired
 };
 
-function toValue(from, to) {
```
JasonStoltz: We used to do some funky stringification of values for the `Select` box. The logic was specific to Range type filters and as it turns out, was 100% unnecessary. We now just pass the value as the `value` to the `Select` and don't worry about what type of value it is.

packages/react-search-ui-views/stories/MultiCheckboxFacet.stories.js
```diff
@@ -27,10 +27,82 @@ const baseProps = {
...
+    count: 11,
+    value: {
+      from: 11,
+      to: 20,
+      name: "The second option"
+    }
+  }
+];
+
+storiesOf("Facets/MultiCheckboxFacet", module)
```
JasonStoltz: Renamed stories to use the "/" syntax which actually nests them in storybook.

storybook

packages/react-search-ui/src/containers/__tests__/Facet.test.js
```diff
@@ -62,33 +71,6 @@ it("should not render a Facet if there are no facet values available", () => {
   expect(wrapper.html()).toEqual(null);
 });
 
-it("should not render a Facet if there are no options available", () => {
```
JasonStoltz: I removed this. This wasn't actually working. Will reevaluate later.

packages/search-ui/src/SearchDriver.js
```diff
@@ -3,6 +3,9 @@ import URLManager from "./URLManager";
 import RequestSequencer from "./RequestSequencer";
 import DebounceManager from "./DebounceManager";
 
+import { adaptFacets } from "./responseAdapter";
```
JasonStoltz: I created these adapters to basically convert from UI State to App Search Query syntax. I'm actually going to move these to be applied in the App Search connector in a subsequent PR.

Right now, connectors convert from App Search query syntax to whatever their API syntax is. I want them to convert directly from UI State to whatever their API syntax is.

packages/search-ui/src/__tests__/actions/addFilter.test.js
```diff
@@ -61,25 +61,41 @@ describe("#addFilter", () => {
 
   it("Adds a new filter", () => {
     expect(
-      subject("test", "value", { initialFilters: [{ initial: ["value"] }] })
-        .filters
-    ).toEqual([{ initial: ["value"] }, { test: ["value"] }]);
+      subject("test", "value", {
+        initialFilters: [{ field: "initial", values: ["value"], type: "all" }]
```
JasonStoltz: A lot of tests had to be updated to simply use the new "filter" syntax in state.

Instead of:
```
[
  {[field]: "value"}
]
```

It's now:
```
[
  {
    field: "field",
    value: "value",
    type: "and" // For future and/or/none filters
  }
]

```

packages/search-ui/src/__tests__/queryString.test.js
```diff
@@ -83,12 +41,6 @@ describe("#parse", () => {
   it("will parse an object correctly", () => {
     expect(subject(stringified)).toEqual(parsed);
   });
-
-  it("will parse an app search filters object correctly", () => {
```
JasonStoltz: Didn't want "App Search" specific tests here. Everything is still covered by other tests.

packages/search-ui/src/requestAdapters.js
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      ...rest
+    };
+  }
+
+  return v;
+}
+
+function rollup(f) {
+  // TODO Move this to the app search connector and just say "app search connector"
```
JasonStoltz: My model says that every type of filter can support nested ORs.

Meaning:
```
{
  field: "field",
  values: ["value"],
  type: "and"
}
```

```
{
  field: "field",
  values: [{from: 1, to: 100}, {from: 900, to: 1000}],
  type: "and"
}
```

Thee above would be valid in my model. However, the App Search API does not support the second case. I think it probably should, but doesn't. I will log an issue to see if we can get support for this.

JasonStoltz: Meaning, I could envision a case where someone facets on a number of ranges, then wants to select multiple ranges within that facet to filter upon.

packages/search-ui/src/responseAdapter.js
```diff
@@ -0,0 +1,38 @@
+function adaptation1AdaptFacetValue(facetValue) {
```
JasonStoltz: Since a lot of the logic in `adaptFacets` just copies the current object in an immutable way, I tried to call out the adaptations in discrete methods, for readability.

packages/react-search-ui-views/package.json
byronhulcher: TIL `This module is around 5 times faster than wrapping assert.deepEqual() in a try/catch.`

packages/search-ui/src/requestAdapters.js
byronhulcher: Just flagging this TODO in here

JasonStoltz: @byronhulcher Thanks, and thanks for reviewing this. I have another PR queued which cleans this up.

Issue #114, opened 2019-02-26
Connectors should accept UI State and convert to API calls


JasonStoltz:
Current, Connectors accept App Search query DSL and convert it to the particular API syntax.

Instead, they should be responsible for receiving a Search UI State and converting it to their particular API syntax.

This is seems cleaner and requires going from Search UI State > App Search Syntax > Another API syntax. It cuts out the middle step.

Issue #99, opened 2019-02-19
Document the Data Model


JasonStoltz:
People will need to know what the data model looks like in order to work with things like filters, facets, and results.

PR #116, opened 2019-02-28
Update Connector interfaces

JasonStoltz:
This PR accomplishes a few tasks:
https://github.com/elastic/search-ui/issues/99 - Document the Data Model
https://github.com/elastic/search-ui/issues/114 - Connectors should accept UI State and convert to API calls
https://github.com/elastic/search-ui/issues/107 - Connector shouldn't have to create ResultItem and ResultList, just particular JSON format

The purpose of it is to simplify Connectors. Connectors now simply accept the current state of the application, and convert that to an API call, and vice versa. This is much simpler than converting to and from App Search API syntax as an intermediate format as we were before.

Additionally, we dropped the usage of the custom ResultItem and ResultList types from the JavaScript api client for additional simplicity.


packages/search-ui-site-search-connector/src/SiteSearchAPIConnector.js
```diff
@@ -36,7 +31,20 @@ function _request(engineKey, method, path, params) {
     }),
     credentials: "include"
   }).then(response => {
-    return response.json();
+    if (response.status === 200) {
+      return response.json();
+    } else {
+      return response
+        .json()
+        .then(json => {
```
byronhulcher: This seems weird... seems like the `.then()` block throws an error, which will always be caught by the `.catch()` block, which will ignore the original `message` content of the Error from the `.then()` block. AFAICT this would NEVER throw an Error with `json.error` as the Error's argument.  It might be clearer to write this out using async/await.

packages/search-ui-site-search-connector/src/requestAdapters.js
```diff
@@ -22,21 +32,59 @@ export function adaptFacetConfig(facets) {
...
+      if (fieldValue.length > 1) {
+        console.warn(
+          "search-ui-site-search-connector: Cannot apply more than 1 none-value filters to a single field"
+        );
+        return acc;
+      }
+
+      const firstValue = fieldValue[0];
+      if (
+        firstValue.from ||
```
byronhulcher: An alternative to this could be `if (Number.isInteger(firstValue.from) || Number.isInteger(.to) {` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger

packages/search-ui-site-search-connector/src/responseAdapter.js
```diff
@@ -0,0 +1,17 @@
...
+  const totalResults = response.info[documentType].total_result_count;
+  const requestId = "";
+  const facets = getFacets(response.info[documentType]);
+
+  return {
+    results,
+    totalPages,
+    totalResults,
+    requestId,
+    ...(Object.keys(facets).length > 0 && { facets })
```
byronhulcher: I _think_ if facets has no keys/is empty then it won't spread anything, so the check for `Object.keys(facets).length` isn't necessary

packages/search-ui/src/SearchDriver.js
```diff
@@ -178,28 +175,11 @@ export default class SearchDriver {
...
-      result_fields: this.result_fields,
-      search_fields: this.search_fields
-    };
-
-    if (sortField && sortDirection) {
-      searchOptions.sort = {
-        [sortField]: sortDirection
-      };
-    }
+    // facets: removeConditionalFacets(
```
byronhulcher: 👀

JasonStoltz: Removed this, it was moved elsewhere and the test is still passing 👍

packages/search-ui-site-search-connector/src/responseAdapter.js
JasonStoltz: Confirmed. So that check is to avoid ending up with:
```
{
  facets: {}
}
```

When it's there, if there is an empty facets object, I simply get:
```
{
}
```

Which is what I want

byronhulcher: I see now that the goal was to include facets as a property under the object only when facets itself had properties, which this code does accomplish!

Issue #105, opened 2019-02-21
Auto complete results


JasonStoltz:
Different than suggestions.

https://github.com/elastic/search-ui/issues/52

Design: https://github.com/elastic/search-ui/issues/113

- [x] Move query configuration to a separate "queries.search" field in Search UI config
- [x] Update setSearchTerm action to include have an "autocomplete" option which invokes additional API call and stores results in state
- [x] Add configuration for "autocomplete" 
- [x] Update SearchBox component to have Autocomplete option
- [x] Create view
- [ ] Document autocomplete | debounce options on components

PR #133, opened 2019-03-08
Added autocomplete results to state

JasonStoltz:
https://github.com/elastic/search-ui/issues/105

- Added an option to `SearchBox` called `autocompleteResults`.
- Added a flag to `setSearchTerm` called `autocompleteResults`.
- Added a new connector method called `autocompleteResults`.
- Added a new state property called `autocompletedResults` (Note the "d").

When the `autocompelteResults` flag is true, a new API request will be invoked to fetch results for and populate `autocompletedResults`.

So basically, if you fire up the sandbox application and start typing, since the `autocompleteResults` flag is set to `true`, you will see API requests being fired off, and the results will actually be populated in state.

This makes them available for us to build a view to consume them, which will be added in a future PR.

latest

### DIFF
--- a/packages/search-ui/src/actions/setSearchTerm.js
+++ b/packages/search-ui/src/actions/setSearchTerm.js
@@ -5,12 +5,13 @@
  *
  * @param searchTerm String
  * @param options Object Additional objects
+ * @param options.autocompleteResults Fetch autocomplete results?
  * @param options.refresh Boolean Refresh search results?
- * @param options.wait Boolean Refresh search results?
+ * @param options.debounce Length to debounce API calls
  */
 export default function setSearchTerm(
   searchTerm,
-  { refresh = true, debounce = 0 } = {}
+  { autocompleteResults = false, refresh = true, debounce = 0 } = {}
 ) {
   this._setState({ searchTerm });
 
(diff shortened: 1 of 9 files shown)


packages/search-ui/src/actions/setSearchTerm.js
```diff
@@ -5,12 +5,13 @@
  *
  * @param searchTerm String
  * @param options Object Additional objects
+ * @param options.autocompleteResults Fetch autocomplete results?
```
goodroot: Can you please create an issue for autocomplete/debounce documentation? 👍

JasonStoltz: Hey @goodroot. I just added a subtask here: https://github.com/elastic/search-ui/issues/105.

I don't mind updating the README / documentation to include that option, but you're certainly welcome to if you'd prefer.

JasonStoltz:
Added an option to SearchBox called autocompleteResults.
Added a flag to setSearchTerm called autocompleteResults.
Added a new connector method called autocompleteResults.
Added a new state property called autocompletedResults (Note the "d").
When the autocompelteResults flag is true, a new API request will be invoked to fetch results for and populate autocompletedResults.

So basically, if you fire up the sandbox application and start typing, since the autocompleteResults flag is set to true, you will see API requests being fired off, and the results will actually be populated in state.

This makes them available for us to build a view to consume them, which will be added in a future PR.

Issue #113, opened 2019-02-26
Proposal for Autocomplete


JasonStoltz:
The term "Autocomplete" refers to the feature, where when text is entered in a search box, some sort of search is performed, and results of which are displayed in a dropdown underneath the search box.

Our Autocomplete can perform two types of completions:

- Results - A search is performed for individual documents based on the current term. The Results of that search are shown, and clicking on a Result will immediately navigate the user to the page corresponding to that result.
- Suggestions - A search for Suggestions is performed. A "Suggestion" is are suggested terms for a query for documents to be performed later. Clicking on a Suggestion will trigger a new search for the selected term.

Configuration can be added for Results and / or Suggestions. Results will be broken into sections, so a "section title" can be added to separate results. More than one type of suggestion can be fetched, suggestion configuration must be keyed by type.

```jsx
<SearchBox
  autocompleteResults={{
    sectionTitle: "Results",
    titleField: "title",
    urlField: "nps_link",
    linkTarget: "_blank",
    shouldTrackClickThrough: true,
    clickThroughTags={['someTag']}
  }}
  autocompleteSuggestions={{
    documents: {
      sectionTitle: "Suggested Queries"
    }
  }}
  autocompleteMinimumCharacters={3}
  autocompleteView={CustomAutocompleteView}
  onSelectAutocomplete{selection => console.log(selection)}
  searchAsYouType={true}
  debounce={100}
/>
```

| option | description |
| ---- | ---- |
| autocompleteResults | Indicates that "results" should be displayed in a dropdown below the search box. The default for this type of autocomplete is that clicking on a result will take you directly to a result detail page. Can be a boolean, or a configuration object containing view configuration. **|
| sectionTitle | Title to show in section within dropdown |
| titleField | Field within a Result to use as the "title" |
| urlField | Field within a Result to use for linking |
| linkTarget | Used to navigate user to a new tab |
| shouldTrackClickThrough | defaults to true |
| clickThroughTags | always sends 'autocomplete', what to send in addition |
| autocompleteQuerySuggestions | Indicates that "query suggestions" should be displayed in a dropdown below the search box. The default for this type of autocomplete is that clicking on a result will populate the search box. Can be a boolean, or a configuration object containing view configuration **|
| searchAsYouType | Existing parameter included to show that this is a separate feature. |
| debounce | |
| autocompleteView | Provide a different view just for the autocomplete dropdown |
| autocompleteMinimumCharacters | Minimum number of characters before autocompleting |
[...]

PR #143, opened 2019-03-25
Full Autocomplete support for Search Results

JasonStoltz:
This PR completes the remaining work for to have a basic working Autocomplete with Search Results.

Design is documented here: https://github.com/elastic/search-ui/issues/113

Remaining work is captured here: https://github.com/elastic/search-ui/pull/140

- Added working example to the Sandbox
- Closes the autocomplete Autocomplete any time a query is submitted
- Ensure the autocomplete value is not changed whenever a result is selected
- Support for boolean configuration of autocompleteResults (explained in design doc)
- Updated `FieldValue` type to expect Arrays of values from the server.

To test this, run the sandbox app and play around with the autocomplete: https://github.com/elastic/search-ui/tree/master/examples/sandbox

latest

To just play around with the view, you can run storybook:

```
cd packages/react-search-ui-views && npm run storybook
```

latest

### DIFF
--- a/packages/react-search-ui-views/src/SearchBox.js
+++ b/packages/react-search-ui-views/src/SearchBox.js
@@ -5,6 +5,16 @@ import Downshift from "downshift";
 import { Result } from "./types";
 import { Suggestion } from "./types";
 
+function getRaw(result, value) {
+  if (!result[value] || !result[value].raw) return;
+  return result[value].raw;
+}
+
+function getSnippet(result, value) {
+  if (!result[value] || !result[value].snippet) return;
+  return result[value].snippet;
+}
+
 function SearchBox(props) {
   const {
     useAutocomplete,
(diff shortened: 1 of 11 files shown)


packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -15,32 +25,48 @@ function SearchBox(props) {
...
-        itemToString={item =>
-          // TODO
-          item
-            ? item[autocompleteResults.titleField]
-              ? item[autocompleteResults.titleField].raw ||
-                item[autocompleteResults.titleField].snippet
-              : item.suggestion
-            : ""
+  const onSelectAutocomplete =
+    props.onSelectAutocomplete ||
```
scottybollinger: You could omit `props.` and add `onSelectAutocomplete` to the destructured props above

packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -15,32 +25,48 @@ function SearchBox(props) {
...
-          return (
+      }
+    });
+
+  return (
+    <Downshift
+      inputValue={value}
+      onChange={onSelectAutocomplete}
+      onInputValueChange={onChange}
+      stateReducer={(state, changes) => {
```
scottybollinger: This could be simplified:
```
stateReducer={(_, changes) => changes}
```

packages/react-search-ui-views/src/SearchBox.js
JasonStoltz: Good suggestion. I did it this was because I am basically trying to default the behavior. I couldn't find a good way to do that with destructuring.

packages/react-search-ui-views/src/SearchBox.js
JasonStoltz: Good catch. That was actually not needed, I just forgot to remove it, so I just removed it.

JasonStoltz:
@scottybollinger Thanks for the feedback, ready for a second look!

PR #170, opened 2019-04-11
Added "debug" option to SearchDriver

JasonStoltz:
You know... for debugging.

latest


examples/sandbox/src/App.js
```diff
@@ -26,6 +26,7 @@ export default function App() {
   return (
     <SearchProvider
       config={{
+        debug: true,
```
omegahm: Debug is default? Is that on purpose?

packages/search-ui/src/actions/trackAutocompleteClickThrough.js
```diff
@@ -8,6 +8,10 @@
  * this click event
  */
 export default function trackAutocompleteClickThrough(documentId, tags = []) {
+  if (this.debug)
+    // eslint-disable-next-line no-console
+    console.log("Action", "trackAutocompleteClickThrough", ...arguments);
```
omegahm: This debug/console is different than all the rest. 🤷🏻‍♂️

examples/sandbox/src/App.js
JasonStoltz: @omegahm Just in the sandbox implementation. It is not on by default otherwise.

packages/search-ui/src/actions/trackAutocompleteClickThrough.js
JasonStoltz: @omegahm It was just because it was automatically wrapped to 2 lines by my editor, due to it's length.

I reformatted it so it looks a bit better.

While I was updating the code, I also changes the state logging to log both the state that changed AND the complete state.

Mind re-approving this?

PR #174, opened 2019-04-12
Save lives, unsubscribe from listeners.

JasonStoltz:
There were errors when using this library in Single Page Applications, like Gatsby, because we never unsubscribed from our listeners. React would attempt to make state updates to a component that was unmounted. This resolves that issue.


packages/react-search-ui/src/SearchProvider.js
```diff
@@ -37,11 +37,19 @@ class SearchProvider extends Component {
     })
   };
 
+  subscription = state => {
+    this.setState(state);
+  };
```
omegahm: This `;` seems out of place? You don't finish the other functions with a semicolon.

JasonStoltz: @omegahm All of semi-colons are managed by Prettier, so It's actually put there by Prettier! 

And in this case it's a "Public Field Definition" as opposed to a method definition like the other functions, so it gets a semi-colon, while the others don't. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Public_field_declarations

PR #175, opened 2019-04-12
Result container component now passes result as a prop to children

### LINKED ISSUE #82: Need access to entire ResultItem in Result view
Currently, raw values are escaped, so there is no way to access the raw value if you actually do want to render some html.

Consider simply passing through the entire ResultItem.

A url field, or image url needs to be raw, for example.

byronhulcher:
With this PR the `Result` React container component passes `result`, `titleField`, and `urlField` from its parent down to its children.  This will enable people passing a custom `view` into `Result` components or using the `renderResult` functionality of `Results` components  to access to raw result data we return from the Swiftype API

Closes #82


packages/react-search-ui/src/containers/Result.js
```diff
@@ -1,18 +1,21 @@
...
 import { Result } from "@elastic/react-search-ui-views";
 
 import { withSearch } from "..";
 import { Result as ResultType } from "../types";
 
-function getRaw(result, value) {
-  if (!result[value] || !result[value].raw) return;
-  return result[value].raw;
+function getFieldType(result, field, type) {
+  if (!result[field] || !result[field][type]) return;
```
omegahm: What happens if the value you want to get back is `false` or `""` or some other falsey value in JS?

packages/react-search-ui/src/containers/Result.js
```diff
@@ -32,16 +42,12 @@ function htmlEscape(str) {
...
-  return Object.keys(result).reduce((acc, n) => {
-    // Fallback to raw values here, because non-string fields
-    // will not have a snippet fallback. Raw values MUST be html escaped.
-    let value = result[n].snippet || htmlEscape(result[n].raw);
-    value = Array.isArray(value) ? value.join(", ") : value;
-    acc[n] = value;
-    return acc;
+function getSafeResult(result) {
+  return Object.keys(result).reduce((acc, field) => {
+    return {...acc, [field]: getSafe(result, field)};
```
omegahm: I believe a linter might tell you to add spaces inside the `{}` here.

packages/react-search-ui/src/containers/Result.js
byronhulcher: It'd return `undefined` I guess.  I'm not opposed to updating this but it was existing code from @JasonStoltz.  Jason, is it good with you if I update to 
```
function getFieldType(result, field, type) {
  if (typeof result[field] == "undefined") return;
  return result[field][type];
}
```

JasonStoltz: I would just return `undefined` here, which I think is OK behavior.

qhoxie: My feeling is that if someone has a value of false or an empty string and they ask for "raw" they should get that value.  It may not be a huge issue in practice but I think it is the right behavior.

packages/react-search-ui/src/containers/Result.js
```diff
@@ -76,14 +82,17 @@ export class ResultContainer extends Component {
     const { result, titleField, urlField, view } = this.props;
     const View = view || Result;
 
+    const safeResult = getSafeResult(result)
+
     return View({
```
qhoxie: I feel like the API we are giving here is a little awkward.  I don't yet have a really strong opinion here, but I'll share my concerns.

1. fields and result serve basically the same purpose.  fields is arguably more friendly but I worry about the ambiguity of mixing raw/snippet/array/joined output and how we explain that to users.
2. safe continues to be a difficult term for me in this context — I'd probably prefer something like "escaped."  safe feels like we are extending our reach into risky areas by making assumptions about how someone is going to use the output.  We are probably right in most cases, though.

A good question to ask it — why do we need fields if we have result?

PR #179, opened 2019-04-12
Added onSubmit

JasonStoltz:
This is primarily so that a SearchBox can be used independently in a header and submit users to the master search page.

Here is an example from a Gatsby site:

```jsx
          <SearchBox
            autocompleteResults={{
              titleField: "title",
              urlField: "nps_link",
            }}
            autocompleteSuggestions={true}
            onSubmit={searchTerm => {
              navigate("/search?q=" + searchTerm)
            }}
            onSelectAutocomplete={(
              selection,
              {},
              defaultOnSelectAutocomplete
            ) => {
              if (selection.suggestion) {
                navigate("/search?q=" + selection.suggestion)
              } else {
                defaultOnSelectAutocomplete(selection)
              }
            }}
          />
```

latest


ADVANCED.md
```diff
@@ -275,17 +275,17 @@ import { SearchBox } from "@elastic/react-search-ui";
...
+| autocompleteView              | Render Function                                                              | no        | [Autocomplete](packages/react-search-ui-views/src/Autocomplete.js) |         | Provide a different view just for the autocomplete dropdown.                                                                                                                                                                                                                                |
+| onSelectAutocomplete          | Function(selection. options, defaultOnSelectAutocomplete)                    | no        |                                                                    |         | Allows overriding behavior when selected, to avoid creating an entirely new view. In addition ot the current `selection`, various helpers are passed as `options` are passed as the second parameter, and the `onSelectAutocomplete` in order to simply wrap and call the default behavior. |
```
goodroot: > In addition ot the current `selection`, various helpers are passed as `options` are passed as the second parameter, and the `onSelectAutocomplete` in order to simply wrap and call the default behavior.

Perhaps: 

> In addition to the current `selection`, various helpers are passed as `options` to the second parameter - and the `onSelectAutoComplete` - to wrap and call the default behaviour.

JasonStoltz: I butchered that wording. How about this?

> In addition to the current `selection`, various helpers are passed as `options` to the second parameter. This third parameter is the default `onSelectAutocomplete`, which allows you to defer to the original behavior.

goodroot: ❤️

PR #192, opened 2019-04-17
Added filter type support

JasonStoltz:
Facet now has a "filterType" option, to configure what sort of a
filter is applied by the Facet selection. To correspond with that
the filter actions have been updated to accept a "filterType" as
well.

```jsx
<SearchProvider config={{
  ...otherConfig,
  searchQuery: {
    disjunctiveFacets: ["states"],
    facets: {
     states: { type: "value", size: 30 }
    }
  }
}}>
  {() => (
    <Facet field="states" label="States" filterType="any" />
  )}
</SearchProvider>
```

latest


allenpansoftware:
This seems not working on ElasticSearch [example ](https://github.com/elastic/search-ui/tree/master/examples/elasticsearch), event I have set the Facet to be "sticky" like this:
<Facet
                      field="states"
                      label="States"
                      filterType="any"
                      isFilterable={true}
                      sticky 
                    />

It still clear out the initial filters, any single filter selection event will trigger the facet to change down into selected filter only. --- could you please advise if this is a designated feature for ElasticSearch sample or there is a way to make it work same as the "Sandbox" sample.

PR #201, opened 2019-04-18
Tag Swiftype requests with "search-ui" header

JasonStoltz:
This PR passes two additional headers back to Site Search an App Search for analytical purposes. These are ONLY passed when using one of these two connectors.

React_App


JasonStoltz:
@goodroot I added a privacy notification to the main README, mind taking a quick look?

PR #203, opened 2019-04-19
Containers no longer hide themselves by default before first search

### LINKED ISSUE #200: Components disappear in certain cases.
In the `Sorting.js` container, we have the following code:

```
if (!searchTerm && results.length === 0) return null;
```

The intent of this code was to hide the Component on the initial load of the Search UI experience, and not show it until it is needed.

There are few issues with this.

1. Individual UI components should not be responsible for this. This is something that should happen at a top level IF it is needed. Not every user of the API will need this. Some users will not have an initial empty state.

2. The check itself is not actually successfully checking for the initial empty state of the UI. It is possible to get back into this case where you have 0 results with no search term by some creative use of filters.

byronhulcher:
Fixes #200

### DIFF
--- a/examples/sandbox/src/App.js
+++ b/examples/sandbox/src/App.js
@@ -114,7 +114,7 @@ export default function App() {
         apiConnector: connector
       }}
     >
-      {_ => (
+      {({ wasSearched }) => (
         <div className="App">
           <ErrorBoundary>
             <Layout
@@ -136,7 +136,7 @@ export default function App() {
               }
               sideContent={
                 <div>
-                  <Sorting
+                  {wasSearched && <Sorting
                     label={"Sort by"}
                     sortOptions={[
                       {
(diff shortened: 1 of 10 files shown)


JasonStoltz:
@byronhulcher Looks good. Let me know if you need help fixing the failing specs.

examples/sandbox/src/App.js
```diff
@@ -173,7 +173,7 @@ export default function App() {
               }
               bodyHeader={
                 <React.Fragment>
-                  <PagingInfo />
+                  {wasSearched && <PagingInfo />}
                   <ResultsPerPage />
```
JasonStoltz: @byronhulcher One small thing since you're in here making updates anyway. Could you update this line to...

```
{wasSearched && <ResultsPerPage />}
```

That should give us a nice clean loading experience. Right now the ResultsPerPage pops in when you start typing text in the sandbox.

latest

PR #204, opened 2019-04-19
Add watch.json file to sandbox example for hotloading

byronhulcher:
I'm using this on one of my Glitch projects, seems like a good idea to add it here?

### DIFF
--- a/examples/sandbox/watch.json
+++ b/examples/sandbox/watch.json
@@ -1,0 +1,19 @@
+{
+  "install": {
+    "include": [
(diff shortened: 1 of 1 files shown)


JasonStoltz:
@byronhulcher Could you explain what this does? The sandbox already has HMR for local development, and I'm not exactly sure what uses this file?

byronhulcher:
Actually I think I'm mistaken, this watch.json thing is specific to Glitch, apologies!

Issue #60, opened 2019-02-05
ResultsPerPage component should have configurable options


JasonStoltz:
Currently, the options "20", "40", and "60" are hardcoded as options in the ResultsPerPage components: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui/src/containers/ResultsPerPage.js#L38.

This should be configurable, as users may want options other than these defaults.

m-sureshraj:
I would like to work on this issue. How can I assign this issue to myself?

JasonStoltz:
@m-sureshraj For now, I'd say just comment here (like you've done) and enter a PR.

m-sureshraj:
PR - https://github.com/elastic/search-ui/pull/224

PR #224, opened 2019-04-28
ResultsPerPage component should have configurable options

m-sureshraj:
## Description
ResultsPerPage component should have configurable options

## List of changes
* Added new prop `resultsPerPageOptions` to the `<ResultsPerPage />` component to configure results per page options.
* Documented `resultsPerPageOptions` prop usage.
* Added test case for`resultsPerPageOptions` 

## Associated Github Issues
#60

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -133,7 +133,7 @@ Request state can be set by:
 | ---------------- | -------------------------------------- | --------- | -------------------------------------- |
 | `current`        | Integer                                | optional  | Current page number                    |
 | `filters`        | Array[[Filter](./src/types/Filter.js)] | optional  |                                        |
-| `resultsPerPage` | Integer                                | optional  | Number of results to show on each page |
+| <a name="resultsPerPageProp"></a>`resultsPerPage` | Integer                                | optional  | Number of results to show on each page |
 | `searchTerm`     | String                                 | optional  | Search terms to search for             |
 | `sortDirection`  | String ["asc" \| "desc"]               | optional  | Direction to sort                      |
 | `sortField`      | String                                 | optional  | Name of field to sort on               |
@@ -408,7 +408,11 @@ import { Results } from "@elastic/react-search-ui";
 
 Shows a dropdown for selecting the number of results to show per page.
 
-Uses 20, 40, 60 as options.
+Uses [20, 40, 60] as a default options. You can use `resultsPerPageOptions` prop to pass custom options. 
+
+**Note:** When passing custom options make sure one of the option value matches 
+the current [`resultsPerPage`](#resultsPerPageProp) value, which is 20 by default.
+To override `resultsPerPage` default value [Refer the custom options example](#Example-using-custom-options).
 
 ### Example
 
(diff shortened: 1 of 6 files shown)


packages/react-search-ui-views/src/ResultsPerPage.js
```diff
@@ -19,6 +19,15 @@ function ResultsPerPage({ className, onChange, options, value }) {
     ? wrapOption(options.find(option => option === selectedValue))
     : null;
 
+  if (selectedOption && !selectedOption.value) {
```
JasonStoltz: I had a thought about this.

Instead of warning when the correct value is not found, I think we should simply show the value.

Something like this:
```
const selectedOption =
    selectedValue && options.includes(selectedValue)
      ? wrapOption(options.find(option => option === selectedValue))
      : wrapOption(selectedValue);

  if (!options.includes(value)) options = [value, ...options];
```

latest

In the example above, the default value is `20`, which is not in the configured list of options, so we simply create an option.

Does that makes sense?

packages/react-search-ui/src/containers/ResultsPerPage.js
```diff
@@ -8,6 +8,7 @@ export class ResultsPerPageContainer extends Component {
     // Props
     className: PropTypes.string,
     view: PropTypes.func,
+    resultsPerPageOptions: PropTypes.arrayOf(PropTypes.number),
```
JasonStoltz: Would it be better to call this `option`? Having a property `resultsPerPageOptions` on a `ResultsPerPage` component sounds redundant.

ADVANCED.md
```diff
@@ -408,7 +408,11 @@ import { Results } from "@elastic/react-search-ui";
 
 Shows a dropdown for selecting the number of results to show per page.
 
-Uses 20, 40, 60 as options.
+Uses [20, 40, 60] as a default options. You can use `resultsPerPageOptions` prop to pass custom options. 
+
+**Note:** When passing custom options make sure one of the option value matches
```
JasonStoltz: Thank you for noting this. 👍

ADVANCED.md
```diff
@@ -408,7 +408,11 @@ import { Results } from "@elastic/react-search-ui";
 
 Shows a dropdown for selecting the number of results to show per page.
 
-Uses 20, 40, 60 as options.
+Uses [20, 40, 60] as a default options. You can use `resultsPerPageOptions` prop to pass custom options. 
+
+**Note:** When passing custom options make sure one of the option value matches 
+the current [`resultsPerPage`](#resultsPerPageProp) value, which is 20 by default.
+To override `resultsPerPage` default value [Refer the custom options example](#Example-using-custom-options).
```
JasonStoltz: Small typos:

`Refer the custom options example` -> `refer to the custom options example`

`the option value matches` -> `the option values match`

packages/react-search-ui/src/containers/ResultsPerPage.js
m-sureshraj: Yes, Make sense.

ADVANCED.md
m-sureshraj: Sorry about that.

packages/react-search-ui-views/src/ResultsPerPage.js
m-sureshraj: Yes, it does. You can see after the first selection, the dynamically added option was removed from the list. The developer may get confused about this behavior. What do you think?

JasonStoltz: That's a fair point.

I believe the responsibility of this component is two-fold:

1. Show the current `resultsPerPage` value
2. Allow a user to change the `resultsPerPage` value

To that end, I do think that it should always show the current resultsPerPage value from state, whether it is a valid option or not.

That being said, that is actually probably outside the scope of this PR. We could address that separately.

m-sureshraj: Sure, Thanks.

m-sureshraj:
@JasonStoltz Can you please review the PR feedback fixes - https://github.com/elastic/search-ui/pull/224/commits/957677150825d28a7a6fc733b6856d337f675222

JasonStoltz:
Thanks again @m-sureshraj !

PR #226, opened 2019-05-03
Updated primary README

JasonStoltz:
Updated some words. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,18 +106,33 @@ Or go "headless", and take complete control over the look and feel of your searc
 </SearchProvider>
 ```
 
-A search experience built with Search UI is made up of four key elements:
+A search experience built with Search UI is composed of the following layers:
 
-1. [A Connector](#1-connector)
-2. [A SearchProvider](#2-searchprovider)
-3. [Components](#3-components)
-4. [Styles and Layout](#4-styles-and-layout)
+1. [A Search API](#1-search-api)
+2. [A Connector](#2-connectors)
+3. [A SearchProvider](#3-searchprovider)
+4. [Components](#4-components)
+5. [Styles and Layout](#5-styles-and-layout)
+
+```
+Styles and Layout -> Components -> SearchProvider -> Connector -> Search API
+```
 
 ---
 
-### 1. Connector
+### 1. Search API
 
-Connectors are modules that tell Search UI how to connect and communicate with a particular API.
+A Search API is any API that you use to search data; examples being Elasticsearch,
+Elastic App Search, and Elastic Site Search.
+
+In order to use Search UI you'll need to have your data indexed into a service like
+this before you can start searching.
+
+### 2. Connectors
+
+Connectors are modules that tell Search UI how to connect and communicate with
+your Search API. They generate Search API calls for you so that Search UI will "just work",
+right out of the box.
 
 Search UI currently provides two Connectors:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,15 +20,15 @@
 
 Search UI allows you to quickly implement search experiences without re-inventing the wheel.
 
-It supports **[React](https://reactjs.org)** and **works with any search API**.
+It is built with **[React](https://reactjs.org)** and **works with any search API**.
```
byronhulcher: Hmmm the components are the only things that use React, right? I feel like saying `Built with React` here might scare people away who might want to build their own vanilla or other framework based components

README.md
```diff
@@ -140,17 +140,10 @@ Search UI can connect to **any** web based Search API. Read [the advanced README
 
 The `SearchProvider` object will tie all of your Components together so that they work as a cohesive application.
 
-It's where you configure your UI, acting as the state manager between "State" and "Actions".
-
-The overall flow is like this:
-
-```
-Components -> SearchProvider -> Connector -> Search API (App Search)
```
byronhulcher: I like this line tbh, I'm going to remake it as a part of my talk because I think it highlights the 3 pieces `search-ui` provides and allows you to emphasize that any of them are able to be swapped out for your own code whenever you like.

JasonStoltz: Ah yes, let me see if I can work that back in at the top of this section.

README.md
JasonStoltz: I don't know what it is about that that I don't like, it just doesn't sound right. "supports React", but it is React! 🤔. I'll revert my change, don't want to confuse people.

byronhulcher: "Includes React components"?

JasonStoltz: I think I'm just overthining it.

README.md
```diff
@@ -136,43 +151,55 @@ const connector = new AppSearchAPIConnector({
...
 
-### 2. SearchProvider
+### 3. SearchProvider
 
-The `SearchProvider` object will tie all of your Components together so that they work as a cohesive application.
+`SearchProvider` is the top level component in your Search UI implementation.
+It is where you configure your search experience, and it ties all of your components together so that they work as a cohesive application.
 
-It's where you configure your UI, acting as the state manager between "State" and "Actions".
+`SearchProvider` can be used with components:
```
goodroot: I love the new content in this section. But I think we can smooth the flow. 🌊 

---

SearchProvider can be used with components:

```
<SearchProvider
  config={{
    apiConnector: connector
  }}
>
  {() => <div className="App">{/* Place Components here! */}</div>}
</SearchProvider>

```

---

The idea that you can use searchProvider without components is important. 

But the fact that you _can_ use it with components should lead into the Component section.

ie.

...

You can use searchProvider directly with state and actions:

```
<SearchProvider
  config={{
    apiConnector: connector
  }}
>
  {({ searchTerm, setSearchTerm }) => (
    <div className="App">{/* Work directly with state and actions! */}</div>
  )}
</SearchProvider>
```

One way to think of `SearchProvider` is as a search experience without a view. Instead of thinking of a search experience in terms of UI like search boxes and dropdowns, it thinks of the underlying "state" and "actions" associated with a search experience. So instead of a "search box", the `SearchProvider` just cares about the underlying `searchTerm` that is populated in a search box (the state), and the action that a user takes on that search box when they change the `searchTerm`, i.e, `setSearchTerm` (the action).

**Since you are able to work directly with "state" and "actions", you can use any type of input you want!** As long as it ends up calling `setSearchTerm`, it will "just work". This gives you maximum flexbility over your experience.

**But there are also pre-built search features consumed by searchProvider. You can quickly add a searchBox, etc, etc. They're called Components.**

_Next section_

---

That way, you introduce that tasty, flexibly awesome searchProvider idea _and then_ the component idea. ❤️

JasonStoltz: I agree, the flow wasn't great before. I adjusted it, but kind of in my own way: https://github.com/elastic/search-ui/pull/226/commits/f87a0582be2cc60433cda03f066eaf7461ddd790.

Let me know what you think.

The verbiage explaining "state" and "actions" more in depth I'm going to move to the Advanced Headless Core guide.

goodroot: 👍

PR #247, opened 2019-05-23
Removed elasticsearch connector

JasonStoltz:
This is to reduce confusion. The connector was experimental and is not recommended for use.


harsimranb:
Are there any plans to add the elastic search connector?

JasonStoltz:
@harsimranb That's a challenging one. I would love to have an Elasticsearch connector in the future, in some form or another.

It's actually a bit of an anti-pattern to connect to an Elasticsearch instance directly from a browser, and even if using a proxy, there are still security concerns around using the full Query DSL.

If there is a connector in the future, it may be more of a server side connector rather than a client side connector. It's still up for debate.

In the meantime, it's entirely possible to use this library with Elasticsearch. We have a little more information on that here: https://github.com/elastic/search-ui#does-search-ui-only-work-with-app-search.

Issue #244, opened 2019-05-21
Add a "useSearch" hook as alternative to "withSearch"


JasonStoltz:
This is just a more modern API

zustand has a good example of what our hook might look like:

https://github.com/react-spring/zustand

m-sureshraj:
@JasonStoltz Can I work on this?

JasonStoltz:
@m-sureshraj Yes! I'm going to publish a PR very soon for performance issues that will relate to this. So I'd say hold off for that PR.

JasonStoltz:
@m-sureshraj https://github.com/elastic/search-ui/pull/249

m-sureshraj:
@JasonStoltz Is there any blockers for this ticket?

JasonStoltz:
@m-sureshraj No blockers, feel free to work on this. I suggest posting the proposed API for this here for discussion before you get too far along on this task, it could help save you some time on the PR if we have the discussion early.

m-sureshraj:
Sure @JasonStoltz, I have not yet started the task. I will post the new API changes here for the discussion soon.

calvinl:
:( 

Would love to see this come back up into active development.

rlueder:
I wrote a custom `useSearch` hook and it works well, I can access any of the `SearchState` values from any other component in my app, examples below:

I created a `Providers.tsx` to wrap my `App` component entry point:

```
import { SearchContext } from "@/hooks/useSearch";

export const Providers = ({ children }: ProvidersProps) => {
  return (
    <SearchProvider config={searchProviderconfig}>
      <WithSearch
        mapContextToProps={(context: SearchContextState) => context}
      >
        {(contextProps: SearchContextState) => (
          <SearchContext.Provider value={contextProps}>
            {children}
          </SearchContext.Provider>
        )}
      </WithSearch>
    </SearchProvider>
  );
};
```
Then `useSearch.ts` hook looks like this:

```
import { createContext, useContext } from "react";

import type { SearchContextState } from "@elastic/search-ui";

export const SearchContext = createContext<SearchContextState | undefined>(
  undefined,
);

export const useSearch = () => {
  const context = useContext(SearchContext);

  if (context === undefined) {
    throw new Error("useSearch must be used within a SearchContextProvider");
  }

  return context;
};

export default useSearch;

```

And this is how it's used:

```
const SomeComponent = () => {
    const { reset, results, resultSearchTerm, totalResults } = useSearch(); // or any other values from SearchContext
    return (
[...]

PR #249, opened 2019-05-28
Performance fixes

### LINKED ISSUE #223: Significant performance issues
There are significant performance issues in a demo use case internally at Elastic.

The setup is Gatsby + Styled components. The render time for `Facet` components on the page seem to be very computationally expensive.

With autocomplete enabled, each facet is rendering twice, for each state update associated with an autocomplete query.

There are few possible things that could being going on here:

1. Facet should be memoizing some sort of computation that it performs.
2. Facet is performing some logic that could be simplified.
3. Facet (and other components) should implement "componentShouldUpdate".
4. There is something inefficient about the interaction between styled components and this library.

JasonStoltz:
- [x] Tests
- [x] Docs
- [x] Test in ecommerce example
- [x] "with" should be "uses"

Feedback fixes:
- [x] Change `withSearch` to use currying
- [x] Change uses parameter in `withSearch` and `SearchConsumer` to be a function that maps context to props
- [x] Rename `uses` parameter to `mapContextToProps`
- [x] Reconcile the existing `mapContextToProps` with the new one ... i.e., the existing one should simply override the default `mapContextToProps` for a component
- [x] `mapContextToProps` should accept a second parameter, which is the current props from the component, which can be used to make decisions in mapping.
- (Edited) `withSearch` should no longer produce a PureComponent, it should work with either. Using a PureComponent or React.Memo should instead be a recommendation in the performance section. All of OUR components should still be pure. EDIT: On second thought, it would be easier to enforce PureComponent for the time being to ensure good performance, and change it later to make it more flexible. It would be harder to go the other way around.
- [x] Rename `SearchConsumer` to `WithSearch`
- [x] Performance section should talk a bit about performance in the same way that https://github.com/drcmda/react-contextual/blob/master/PITFALLS.md does.
- [x] Verify that `WithSearch` HOC does not remount children on every render.

## Description

This PR fixes performance issues with this library, specifically related to over-rendering, as described in the React documentation [here](https://reactjs.org/docs/optimizing-performance.html#avoid-reconciliation).

The issue is best illustrated by the following GIF:

latest

That GIF was made using the React plugin for Chrome, which will highlight sections of the page that update. (More on that [here](https://reactjs.org/docs/optimizing-performance.html#avoid-reconciliation)).

As you can see in the GIF, the entire experience is updated on every single state update.

For simple sites, the performance of this is still reasonable. However, for some applications which do more on the front end, the re-renders can be come quite noticeable. For example: http://public-demo-assets.swiftype.info/gallivant/public/search/?q=shor&size=n_20_n.

latest

The performance here is quite bad. There is a noticeable delay when attempting to search on that page. And as you can see, the entire page is updating as you type.

There are 2 key contributing factors to this performance issues:
1. State for the entire experience is managed in the top-level `Provider` component, which causes full sub-tree re-renders whenever a state change occurs.
2. Individual components receive ALL state as props, so they are re-rendered with every state update.

Performanc_key1
                                
The solutions involved two things:
1. Subscribe to and manage state at the component level, rather than the top level, to avoid full sub-tree re-renders.
2. At the component level, make components "Pure Components", and have them only accept props that are relevant to that component. In other words, make it so that components only update when pieces of the state tree that they care about are updated.

Performanc_key2

## List of changes

In order to properly implement this, I made the following breaking changes to the API:

**withSearch now accepts a mapContextToProps function as the first parameter**

```
export default withSearch(({ results }) => ({ results }))(ResultsContainer);
```

In that function, you must specify which actions and state your component will use. They will be passed as props to that component. Additionally, that component will be a pure component, so it will only update when those props are changed.

**components are now pure**

As mentioned in the previous comment, components are now "pure". That means, when working with components, you must be certain of a couple of things:

There are a couple of caveats that come along with this, which are described here: https://reactjs.org/docs/optimizing-performance.html#examples and in the Performance section of this README>

**SearchProvider no longer accepts a render prop**

You now must use a `WithSearch` component to access state and actions directly as a render prop

Before:
```
<SearchProvider config={config}>
  {({ searchTerm, setSearchTerm, results }) => {
    return (
      <div>
        <input
          value={searchTerm}
          onChange={e => setSearchTerm(e.target.value)}
        />
        {results.map(r => (
          <div key={r.id.raw}>{r.title.raw}</div>
        ))}
      </div>
    );
  }}
</SearchProvider>
```

After:
```
<SearchProvider config={config}>
  <WithSearch
      mapContextToProps={({ searchTerm, setSearchTerm, results }) => ({
        searchTerm,
        setSearchTerm,
        results
      })}
    >
    {({ searchTerm, setSearchTerm, results }) => {
      return (
        <div>
          <input
            value={searchTerm}
            onChange={e => setSearchTerm(e.target.value)}
          />
          {results.map(r => (
            <div key={r.id.raw}>{r.title.raw}</div>
          ))}
        </div>
      );
    }}
  </SearchConsumer>
</SearchProvider>
```

**Stylistic changes**

- The `withSearch` Higher Order component changed to more closely follow patterns set forth by other libraries, like Redux's `connect` and Material UI's `withStyles`. Namely, the `mapContextToProps` pattern closely resembles `mapStateToProps` from Redux, and it uses a curried function.

```
export default withSearch(({ results }) => ({ results }))(ResultsContainer);
```

- The `SearchConsumer` component is now `WithSearch`, to indicate that is primarily the same thing as `withSearch`, but using a Render Props pattern rather than a Higher Order Component.

**React version dependency**

Updated the react dependency version to 16.8 to allow us to use hooks.

## Associated Github Issues

Fixes #223


packages/react-search-ui/src/SearchProvider.js
```diff
@@ -5,35 +5,34 @@ import { SearchDriver } from "@elastic/search-ui";
...
+  constructor() {
+    super();
+    this.state = {
+      driver: null
+    };
+  }
 
   componentDidMount() {
     const { config } = this.props;
     this.driver = new SearchDriver(config);
```
m-sureshraj: Could you explain why the initialization not inside the constructor?
```js
this.state = {
  driver: new SearchDriver(config)
};
```

packages/react-search-ui/src/SearchProvider.js
```diff
@@ -42,16 +41,16 @@ class SearchProvider extends Component {
...
+    if (!this.state.driver) return null;
 
     const providerValue = {
-      ...this.state,
-      ...this.driver.getActions()
+      driver: this.state.driver
     };
 
     return (
       <SearchContext.Provider value={providerValue}>
```
m-sureshraj: This line equivalent to `<SearchContext.Provider value={{ driver: this.state.driver }}>`. 

**From the react doc**
> Because context uses reference identity to determine when to re-render, there are some gotchas that could trigger unintentional renders in consumers when a provider’s parent re-renders. For example, the code below will re-render all consumers every time the Provider re-renders because a new object is always created for value.

https://reactjs.org/docs/context.html#caveats

**More information about this**
* https://medium.com/@ryanflorence/react-context-and-re-renders-react-take-the-wheel-cd1d20663647
* https://kentcdodds.com/blog/how-to-optimize-your-context-value

**Solution 1**
```js
// pseudocode

class SearchProvider extends Component {
  constructor(props) {
    super(props);
    this.state = {
      driver: new SearchDriver(props.config)
    };
  }

  componentWillUnmount() {
    this.state.driver.tearDown();
  }

  render() {
    if (!this.state.driver) return null; // do we need this line?

   return (
      <SearchContext.Provider value={this.state.driver}>
         {this.props.children}
      </SearchContext.Provider>
   );
  }
}
```
**Solution 2 with hooks**
```js
import React, { useState, useEffect } from 'react';

function SearchProvider(props) {
  const [driver,] = useState(() => new SearchDriver(props.config));

  useEffect(() => driver.tearDown, []);

  if (!driver) return null; // Do we need this line?
  
  return <SearchContext.Provider value={driver} {...props} />
}
```
Let me know what do you think

packages/react-search-ui/src/withSearch.js
```diff
@@ -1,23 +1,81 @@
...
-  };
+export default function withSearch(uses = [], Component) {
+  class WithSearch extends React.PureComponent {
+    constructor() {
+      super();
+      this.state = {};
+      this.uses = uses;
+    }
+
+    componentWillMount() {
```
m-sureshraj: componentWillMount is deprecated and will be removed - https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

packages/react-search-ui/src/SearchProvider.js
JasonStoltz: @m-sureshraj I thought I posted a reply to this, but it looks like I did not, so rewriting it.

This is in CDM because `SearchDriver` has a dependency right now on `window`. So using this with SSR in a project like Gatsby will error out. By moving this to CDM we are basically avoiding the problem by simply not trying to instantiate it during SSR.

packages/react-search-ui/src/withSearch.js
JasonStoltz: Good catch.

packages/react-search-ui/src/SearchProvider.js
JasonStoltz: @m-sureshraj Updated, thanks. I did not move this to hooks, since I was unsure of the behavior of `useEffect` in SSR.

yakhinvadim:
Hey Jason! I have some thoughts on this, but they are not strong opinions since I don't know the code nearly as good as you.

I like your solution, although I'm not sure about the API changes.

I think using strings could be inconvenient for static code checking (TypeScript, for example). For example if we rename a variable, Typescript could find all it's usages, but I'm not sure if it could find it in the string.

### HOC

I think the better way of passing a list of properties to component would be to have some kind of mapContextToProps function. Also, Component could be passed not as a second argument of a first call, but as a first argument of a second call. Here's how it could look like:

Instead of:
```
withSearch(["filters", "facets"], Component);
```

It could be:
```
withSearch(context => ({filters: context.filters, facets: context.facets}))(Component)
```

That could help in a couple of ways:
1) It would solve the issue with static type checkers (described above)
2) It would make a function look similar to `connect` from react-redux which I expect many people in React community is used to (see https://react-redux.js.org/api/connect#example-usage)  or `withStyles` from MaterialUI (see https://material-ui.com/styles/api/#withstyles-styles-options-higher-order-component) 

In order to pass all the properties to the component we would need to do this:
```
withSearch(context => context)(Component).
```
But it's not what users should do anyway since it's not good for performance.

### Render Props 
I'm not sure how it should look like for SearchConsumer though. I haven't used render props approach before. I guess instead of:

```
<SearchConsumer with={["filters", "facets"]}>
    {({  filters, facets }) => {
      return (
        <div>
          ...
        </div>
      );
    }}
  </SearchConsumer>
```

it could look like this:

```
<SearchConsumer mapContextToProps={
context => ({filters: context.filters, facets: context.facets})
}>
    {({ filters, facets }) => {
      return (
        <div>
          ...
        </div>
      );
    }}
  </SearchConsumer>
```

but I'm not sure if I like it or not.

I have some thoughts on this but they're kind of offtopic:

### Offtopic
As far as I understand there are three common approaches to extract logic in react: HOCs, Render Props and Custom hooks. 

I think we will need to support hooks eventually. It could look something like this: 
```
const CustomSearchComponent = () => {
  const {facets, filters} = useSearch();
  return ...
}
```

We will also need to support either HOCs or Render Props for those who use classes. But I'm not sure that we need to support both. Maybe while we're still in beta we could drop one of them. That would give us more time to work on other features and would not take much from the users.

Please let me know what you think. As I said, you know the codebase an order of magnitude better than me, so I might be very wrong here.

Also I'm continuing looking on PR and will let you know if I find something else.

ADVANCED.md
```diff
@@ -1260,6 +1237,43 @@ need to have in common is an `additionalOptions` parameter.
...
+The thing to be most cautious of is not to
+[mutate state](https://reactjs.org/docs/optimizing-performance.html#the-power-of-not-mutating-data) that you will
+pass as props to any of these components.
+
+Example of what not to do:
+
+```jsx
+class SomeComponent extends React.Component {
+  changeSorting = () => {
+    const { options } = this.state.options;
```
yakhinvadim:
```suggestion
    const { options } = this.state;
```

ADVANCED.md
```diff
@@ -1260,6 +1237,43 @@ need to have in common is an `additionalOptions` parameter.
...
+    const { options } = this.state;
+    return <Sorting options={options} />;
+  }
+}
+```
+
+Instead, do:
+
+```jsx
+const { options } = this.state.options;
```
yakhinvadim:
```suggestion
const { options } = this.state;
```

yakhinvadim:
```suggestion
```

ADVANCED.md
```diff
@@ -1260,6 +1237,43 @@ need to have in common is an `additionalOptions` parameter.
...
+this.setState({ options: [...options, "newOption"] });
```
yakhinvadim:
```suggestion
this.setState(prevState => ({ options: [...prevState.options, "newOption"] }));
```

yakhinvadim: Using prevState argument instead of previously saved state ensures that we work with the current state.

packages/react-search-ui/src/SearchProvider.js
```diff
@@ -1,57 +1,61 @@
...
+      driver: null
+    };
+  }
 
   componentDidMount() {
     const { config } = this.props;
-    this.driver = new SearchDriver(config);
-    this.setState(this.driver.getState());
-    this.driver.subscribeToStateChanges(this.subscription);
+    // This initialization is done inside of componentDidMount, because initializing the SearchDriver server side
```
yakhinvadim: Just a nitpick:
I think it would be more concise and clear to check if the have `window` in the constructor:

```
const driver = typeof window !== 'undefined' ? new SearchDriver(config) : null;
this.setState({
      driver
    });
```

packages/react-search-ui/src/__tests__/SearchConsumer.test.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { mount } from "enzyme";
+
+import { SearchProvider, SearchConsumer } from "../..";
+
+describe("SearchConsumer", () => {
+  it("exposes state and actions to components", () => {
```
yakhinvadim: Should we test that we expose actions or is testing that we expose state is enough?

packages/react-search-ui/src/__tests__/SearchProvider.test.js
```diff
@@ -1,3 +1,28 @@
-it("needs tests", () => {
-  expect(true).toBe(true);
+import React from "react";
+import { mount } from "enzyme";
+
+import { SearchProvider, SearchConsumer } from "../..";
+
+describe("SearchProvider", () => {
+  it("exposes state and actions to components", () => {
```
yakhinvadim: Same question here. Should we test that we expose actions or is testing that we expose state is enough?

JasonStoltz:
@yakhinvadim 

## HOC

> I think using strings could be inconvenient for static code checking (TypeScript, for example). For example if we rename a variable, Typescript could find all it's usages, but I'm not sure if it could find it in the string.

I haven't used TS. How big of an issue do you perceive this to be? Is it an inconvenience for users of the library, or for us, the maintainers?

> withSearch(context => ({filters: context.filters, facets: context.facets}))(Component)

I personally prefer the current syntax: `withSearch(["filters", "facets"], Component);`. I found the concept of `mapStateToProps` a bit confusing when I first stated using Redux, and I think this syntax is just a bit simpler. Up for debate though, that's just my preference.

My hope was actually that the pattern I put forward is somewhat similar to the pattern of `useEffect`'s second parameter.

```
useEffect(() => {}, [count]); // Only re-run the effect if count changes
```

I'm open to changing it if we all feel there is a very clear, strong benefit in doing so.

## Render Props / Offtopic

I feel like maybe `withSearch` (HOC) and `SearchConsumer` (Render Props) have different use cases.

`SearchConsumer` is nice for inlining some search dependent code without creating a new component: https://github.com/JasonStoltz/search-ui/blob/performance-spike/ADVANCED.md#2-searchconsumer

And `withSearch` is nice for creating custom components: https://github.com/JasonStoltz/search-ui/blob/performance-spike/ADVANCED.md#build-your-own-component

SearchConsumer under the hood just wraps `withSearch` anyway, so it'd not much additional work to maintain: https://github.com/JasonStoltz/search-ui/blob/performance-spike/packages/react-search-ui/src/SearchConsumer.js.

I also agree that we should definitely add a `useSearch` hook: https://github.com/elastic/search-ui/issues/244.

packages/react-search-ui/src/SearchProvider.js
JasonStoltz: @yakhinvadim I worry then, what would happen if we use this in SSR, like in Gatsby. Will the driver **ever** be initialized?

packages/react-search-ui/src/__tests__/SearchConsumer.test.js
JasonStoltz: Good point, I think we should test this.

m-sureshraj:
@JasonStoltz 

>  I personally prefer the current syntax: withSearch(["filters", "facets"], Component);. I found the concept of mapStateToProps a bit confusing when I first started using Redux, and I think this syntax is just a bit simpler. Up for debate though, that's just my preference.

Yes, I agree with this point but we already have an issue #244 to replace `withSearch` with `useSearch` hook, so I think the current `withSearch` syntax will be removed right?

JasonStoltz:
> @JasonStoltz
> 
> > I personally prefer the current syntax: withSearch(["filters", "facets"], Component);. I found the concept of mapStateToProps a bit confusing when I first started using Redux, and I think this syntax is just a bit simpler. Up for debate though, that's just my preference.
> 
> Yes, I agree with this point but we already have an issue #244 to replace `withSearch` with `useSearch` hook, so I think the current `withSearch` syntax will be removed right?

I'd actually like to keep both. Since not everyone is using hooks, I'd like to be able to provide both for users.

packages/react-search-ui/src/SearchProvider.js
yakhinvadim: We had a call with Jason and I agreed that we shouldn't risk not having driver initialized for slightly less code.

yakhinvadim:
We had a call with Jason and decided that it's better to have a syntax that is familiar to React community and supports static typechecking with a cost of having a slightly more complex syntax. Also the syntax was improved: 
Previous suggestions:
```
withSearch(["filters", "facets"], Component);
```
```
withSearch(context => ({filters: context.filters, facets: context.facets}))(Component)
```

Current syntax:
```
withSearch(({filters, facets}) => ({filters, facets}))(Component)
```
The function `({filters, facets}) => ({filters, facets})` would be called mapContextToProps. We are not calling it mapStateToProps to set it apart from redux's mapStateToProps which only works with state and doesn't work with actions. 
It will serve two purposes:
1) performance optimizations (which context updates should rerender the component)
2) state modification (as described here https://github.com/elastic/search-ui/blob/master/ADVANCED.md#mapcontexttoprops)

packages/react-search-ui/src/__tests__/withSearch.test.js
```diff
@@ -122,33 +125,33 @@ describe("withSearch", () => {
...
     it("will not inject an  unspecified state", () => {
-      const element = setup(["resultSearchTerm"]);
+      const element = setup(({ resultSearchTerm }) => ({ resultSearchTerm }));
 
       expect(element.text()).toEqual("another search term");
     });
 
-    it("will inject everything if an empty array is passed", () => {
-      const element = setup([]);
+    it("will inject everything if nothing is passed", () => {
```
yakhinvadim: I don't have a strong opinion on this, so if you disagree, I'll back up.

I think that it would be better to follow the "performance by default" principle here, i.e. if you don't pass a list of props you're using, the component won't update.
The "poor performance" way could be opt-in with passing the entire state deliberately `withSearch(context => context)(Component)`.

I think if we go that way, the search-ui could avoid getting a reputation as a laggy library from those who try to setup a library as fast as possible and didn't read the docs carefully.

Also, if nothing is passed, we could give a console warn like "please provide a list of props".

JasonStoltz: @yakhinvadim That's a fair point. It's also nice as we talked about before to have just 1 way to do things. One less decision to make, you just always pass a `mapContextToProps` function.

JasonStoltz: One less thing to explain.

JasonStoltz: @yakhinvadim https://github.com/elastic/search-ui/pull/249/commits/3d2b933700e1106ccc5fff6ec49d4cc213fde30f

yakhinvadim: Great!

packages/react-search-ui/src/__tests__/SearchConsumer.test.js
JasonStoltz: I have a `will inject a specified action` test for this.

JasonStoltz: Done.

ADVANCED.md
```diff
@@ -75,76 +76,76 @@ It looks like this:
...
-    </div>
-  )}
-</SearchProvider>
-```
-
-You could also use it to add a clear filters button:
+// Selects `searchTerm` and `setSearchTerm` for use in Component
+withSearch(
+  ({ searchTerm, setSearchTerm }) => ({ searchTerm, setSearchTerm }),
+  Component
```
yakhinvadim: Shouldn't it be curried?

ADVANCED.md
```diff
@@ -75,76 +76,76 @@ It looks like this:
...
-      )}
-    </div>
-  )}
-</SearchProvider>
+// Uses current `props` to conditionally modify context
+withSearch(
+  ({ searchTerm }, { someProp }) => ({
+    searchTerm: someProp ? "" : searchTerm
+  }),
+  Component
```
yakhinvadim: Shouldn't it be curried?

ADVANCED.md
```diff
@@ -1131,10 +1136,18 @@ function ClearFilters({ filters, clearFilters }) {
...
 
-export default withSearch(ClearFilters);
+export default withSearch(({ filters, clearFilters }) => ({
+  filters,
+  clearFilters
+}))(ClearFilters);
 ```
 
-## Connectors and Handlers
+Note that `withSearch` accepts an array as the first parameter. Use this property to pass a function that picks which action and state properties you want to use from the Context.
```
yakhinvadim: I'm not sure it accepts an array now.

README.md
```diff
@@ -147,7 +153,7 @@ Search UI currently provides two Connectors:
 1. **Elastic App Search**: [search-ui-app-search-connector](packages/search-ui-app-search-connector)
 2. **Elastic Site Search**: [search-ui-site-search-connector](packages/search-ui-site-search-connector)
 
-The [example search experience](#search-ui) above uses the Elastic App Search Connector:
+The [example search experience](#search-ui) above mapContextToProps the Elastic App Search Connector:
```
yakhinvadim: I think this shouldn't have been replaced :)

packages/react-search-ui/src/WithSearchRenderProps.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import withSearch from "./withSearch";
+import PropTypes from "prop-types";
+
+function WithSearch({ mapContextToProps: mapContextToProps, children }) {
```
yakhinvadim: What does this part mean: "mapContextToProps: mapContextToProps"?

ADVANCED.md
JasonStoltz: Yep, missed that, thank you.

README.md
JasonStoltz: 🤦‍♂

packages/react-search-ui/src/WithSearchRenderProps.js
JasonStoltz: @yakhinvadim Another find / replace typo, something my editor did, will clean it up and double check for any more errors. TY.

PR #258, opened 2019-06-10
README Update

goodroot:
Streamlines things to assume App Search as the underlying Search API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,6 @@ _Read the [advanced README](./ADVANCED.md#build-your-own-connector) to learn how
 
 ### 3. SearchProvider
 
-### 3. SearchProvider
-
 `SearchProvider` is the top level component in your Search UI implementation.
 
 It is where you configure your search experience and it ties all of your components together, so that they work as a cohesive application.
@@ -175,9 +173,9 @@ It is where you configure your search experience and it ties all of your compone
 
 While components can be handy, your search experience sometimes has requirements that don't quite fit what components provide "out of the box".
 
-You can work directly with the "actions" and "state" provided by something we call the "SearchDriver".
+You can work directly with the "actions" and "state" using the "SearchDriver".
 
-Use `WithSearch` to access those in a [Render Prop](https://reactjs.org/docs/render-props.html), giving you maximum flexibility over the experience.
+Use `WithSearch` to access "actions" and "state" in a [Render Prop](https://reactjs.org/docs/render-props.html), giving you maximum flexibility over the experience.
 
 ```jsx
 <SearchProvider
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 Search UI allows you to quickly implement search experiences without re-inventing the wheel.
 
-It supports **[React](https://reactjs.org)** and **works with any search API**.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,7 @@
 
 Search UI allows you to quickly implement search experiences without re-inventing the wheel.
 
-It supports **[React](https://reactjs.org)** and **works with any search API**.
+Supports **[React](https://reactjs.org)** and **works Elastic App Search, or any search API**.
```
JasonStoltz:
```suggestion
Supports **[React](https://reactjs.org)** and **works with Elastic App Search, or any search API**.
```

JasonStoltz: General note, this sounds wishy-washy. "Supports" React, "works with" Elastic App Search, "or" any search API.

What if we could say something more assertive, like:

"A React library for Elastic App Search."

or

"A React library for Elastic App Search that allows you to quickly implement search experiences without re-inventing the wheel"

Just thinking out loud.

README.md
```diff
@@ -136,24 +129,21 @@ Styles and Layout -> Components -> SearchProvider -> Connector -> Search API
...
+You can start a [14 day trial of the managed service](https://www.elastic.co/cloud/app-search-service) or [host the self managed package for free](https://www.elastic.co/downloads/app-search).
 
-Search UI currently provides two Connectors:
+Once your data is indexed into App Search or a similar service, you're good to go.
+
+### 2. Connectors
 
-1. **Elastic App Search**: [search-ui-app-search-connector](packages/search-ui-app-search-connector)
-2. **Elastic Site Search**: [search-ui-site-search-connector](packages/search-ui-site-search-connector)
+A connector is a module that tell Search UI how to connect and communicate with your Search API.
```
JasonStoltz:
```suggestion
A connector is a module that tells Search UI how to connect and communicate with your Search API.
```

README.md
```diff
@@ -180,9 +171,7 @@ It is where you configure your search experience, and it ties all of your compon
 </SearchProvider>
 ```
 
-For more on components, continue to the next section!
-
-While components can be handy, your search experience sometimes has requirements that don't quite fit what components provide "out of the box". In this case, it can be convenient to work directly with the "actions" and "state" provided by something we call the "SearchDriver". You can use `WithSearch` to access those in a [Render Prop](https://reactjs.org/docs/render-props.html), which gives you maximum flexibility over your experience.
+While components can be handy, a search experience can have requirements that don't quite fit what components provide "out of the box". Use `WithSearch` to access "actions" and "state" in a [Render Prop](https://reactjs.org/docs/render-props.html), giving you maximum flexibility over the experience.
```
JasonStoltz: Side note for another time. In our docs we suggest that you can use "Components" or "work with state and actions" directly. We also refer to "state and actions" as "the headless core" in places.

What do you think about eventually calling "state and actions" / "the headless core" simply "the API", or "the low-level API".

I think it would simplify the messaging. And "low level API" is easier to wrap your head around than "headless core".

goodroot: I think so, too -- "core" has taken on a few meanings. I am in agreement.

README.md
goodroot: Agreed! Went with option b, hyperlinking React and App Search.

PR #260, opened 2019-06-11
Added Geo Facet filter support

### LINKED ISSUE #33: When using dynamic date range filters, reloading the page will not correctly select a Date Range in Facet options.
For example, if you select a "Last 30 Minutes" filter, and reload the page, the option that is generate for the last 30 days will not match what is stored in the URL. This causes the filter go be applied, but not show up as a selection in the Facet Filter.

Possible solutions:
- Have a new component, called "Filter Breadcrumb", which gives you breadcrumbs for all applied filters and lets you selectively remove them.
- If a value is applied, but no option is found in a Facet filter, then generate a new option and render it, for the currently applied value.

### LINKED ISSUE #54: Support for Geo facet filters
Currently, if you try to use a Facet on a Geo field, it will not work.

JasonStoltz:
## Description

Geo "facets" were already supported, but actually clicking on filters
would cause an API error in App Search.

This is because we derive the filter values for fields based on the
data in the API response for that facet. The problem with that is,
the API response does not return in the response what the original
`center` and `unit` values were. So in order to build a filter from
a facet, we need to hold on to the requested `center` and `unit` values
from the facet request so we can feed them back to the API in filters.

Ex: the original facet API request, which contains a `center` and
`unit`:

```
facets: {
  location: {
    center: "73.102, -78.120",
    type: "range",
    unit: "u",
    ranges: [
      { from: 0, to: 100, name: "Nearby" },
      { from: 100, to: 500, name: "A longer drive" }
    ]
  }
}
```

Ex response, which does not have `unit` or `center`:
```
facets: {
  location: [
    {
      type: "range",
      data: [
        {
          to: 100,
          from: 0,
          name: "Nearby",
          count: 0
        },
        {
          to: 500,
          from: 100,
          name: "A longer drive.",
          count: 4
        }
      ]
    }
  ]
}
```

We now inject the originally requested `unit` and `center` values into
the resulting application state (done in AppSearchAPIConnector /
responseAdapter).
```
{
  type: "range",
  field: "location",
  data: [
    {
      count: 0,
      value: {
        center: "73.102, -78.120",
        unit: "u",
        to: 100,
        from: 0,
        name: "Nearby"
      }
    },
    {
      count: 4,
      value: {
        center: "73.102, -78.120",
        unit: "u",
        to: 500,
        from: 100,
        name: "A longer drive."
      }
    }
  ]
}
```

Which we can the pass on to the API in a subsequent request as a
filter:
```
filters: {
  "all": [
    {
      "any": [
        {
          "location": {
            "to": 500,
            "from": 100,
            "unit": "mi",
            "center": "40.037461, -76.30460889999999"
          }
        }
      ]
    }
  ]
}
```

I've additionally updated the demo to include a filter for "Distance" and "Date Established. Distance to show Geo, and Date Established to cover dynamic date ranges, which relates to the bug linked below.

latest

## List of changes

- Added support for Geo facet filtering
- Updated demo to include Geo facet filtering
- Fixed #33 as part of this, because it suffered from the same issue.

## Associated Github Issues

Fixes #54 
Fixes #33

### DIFF
--- a/examples/sandbox/src/App.js
+++ b/examples/sandbox/src/App.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import AppSearchAPIConnector from "@elastic/search-ui-app-search-connector";
 import SiteSearchAPIConnector from "@elastic/search-ui-site-search-connector";
 import {
@@ -52,7 +52,7 @@ if (process.env.REACT_APP_SOURCE === "SITE_SEARCH") {
   });
 }
 
-const config = {
+const getConfig = position => ({
   debug: true,
   searchQuery: {
     result_fields: {
@@ -72,10 +72,20 @@ const config = {
         }
       }
     },
-    disjunctiveFacets: ["acres", "states"],
+    disjunctiveFacets: ["acres", "states", "location"],
     facets: {
       world_heritage_site: { type: "value" },
       states: { type: "value", size: 30 },
+      location: {
+        center: position,
+        type: "range",
+        unit: "mi",
+        ranges: [
+          { from: 0, to: 100, name: "Nearby" },
+          { from: 100, to: 500, name: "A longer drive" },
+          { from: 500, name: "Perhaps fly?" }
+        ]
+      },
       acres: {
         type: "range",
         ranges: [
@@ -124,14 +134,35 @@ const config = {
     }
   },
   apiConnector: connector
-};
+});
 
 export default function App() {
+  const [position, setPosition] = useState("");
+  useEffect(
+    () => {
+      if ("geolocation" in navigator) {
+        navigator.geolocation.getCurrentPosition(
+          position => {
+            setPosition(
+              `${position.coords.latitude}, ${position.coords.longitude}`
+            );
+          },
+          () => {
+            setPosition("37.7749, -122.4194");
+          }
+        );
+      } else {
+        setPosition("37.7749, -122.4194");
+      }
+    },
+    [position]
+  );
+
+  if (!position) return null;
+
   return (
-    <SearchProvider config={config}>
-      <WithSearch
-        mapContextToProps={({ wasSearched }) => ({ wasSearched })}
-      >
+    <SearchProvider config={getConfig(position)}>
+      <WithSearch mapContextToProps={({ wasSearched }) => ({ wasSearched })}>
         {({ wasSearched }) => {
           return (
             <div className="App">
@@ -171,6 +202,11 @@ export default function App() {
                         field="visitors"
                         label="Visitors"
                         view={SingleLinksFacet}
+                      />
+                      <Facet
+                        field="location"
+                        label="Distance"
+                        filterType="any"
                       />
                       <Facet
                         field="acres"
--- a/examples/sandbox/src/App.js
+++ b/examples/sandbox/src/App.js
@@ -52,7 +52,7 @@ if (process.env.REACT_APP_SOURCE === "SITE_SEARCH") {
   });
 }
 
-const getConfig = position => ({
+const config = {
   debug: true,
   searchQuery: {
     result_fields: {
@@ -72,20 +72,10 @@ const getConfig = position => ({
         }
       }
     },
-    disjunctiveFacets: ["acres", "states", "location"],
+    disjunctiveFacets: ["acres", "states"],
     facets: {
       world_heritage_site: { type: "value" },
       states: { type: "value", size: 30 },
-      location: {
-        center: position,
-        type: "range",
-        unit: "mi",
-        ranges: [
-          { from: 0, to: 100, name: "Nearby" },
-          { from: 100, to: 500, name: "A longer drive" },
-          { from: 500, name: "Perhaps fly?" }
-        ]
-      },
       acres: {
         type: "range",
         ranges: [
@@ -93,6 +83,17 @@ const getConfig = position => ({
           { from: 0, to: 1000, name: "Small" },
           { from: 1001, to: 100000, name: "Medium" },
           { from: 100001, name: "Large" }
+        ]
+      },
+      location: {
+        // San Francisco. In the future, make this the user's current position
+        center: "37.7749, -122.4194",
+        type: "range",
+        unit: "mi",
+        ranges: [
+          { from: 0, to: 100, name: "Nearby" },
+          { from: 100, to: 500, name: "A longer drive" },
+          { from: 500, name: "Perhaps fly?" }
         ]
       },
       visitors: {
@@ -134,34 +135,11 @@ const getConfig = position => ({
     }
   },
   apiConnector: connector
-});
+};
 
 export default function App() {
-  const [position, setPosition] = useState("");
-  useEffect(
-    () => {
-      if ("geolocation" in navigator) {
-        navigator.geolocation.getCurrentPosition(
-          position => {
-            setPosition(
-              `${position.coords.latitude}, ${position.coords.longitude}`
-            );
-          },
-          () => {
-            setPosition("37.7749, -122.4194");
-          }
-        );
-      } else {
-        setPosition("37.7749, -122.4194");
-      }
-    },
-    [position]
-  );
-
-  if (!position) return null;
-
   return (
-    <SearchProvider config={getConfig(position)}>
+    <SearchProvider config={config}>
       <WithSearch mapContextToProps={({ wasSearched }) => ({ wasSearched })}>
         {({ wasSearched }) => {
           return (
--- a/examples/sandbox/src/App.js
+++ b/examples/sandbox/src/App.js
@@ -215,7 +215,11 @@ export default function App() {
                         label="Date Established"
                         filterType="any"
                       />
-                      <Facet field="location" label="Distance" />
+                      <Facet
+                        field="location"
+                        label="Distance"
+                        filterType="any"
+                      />
                       <Facet
                         field="acres"
                         label="Acres"
--- a/examples/sandbox/src/App.js
+++ b/examples/sandbox/src/App.js
@@ -1,4 +1,6 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
+import moment from "moment";
+
 import AppSearchAPIConnector from "@elastic/search-ui-app-search-connector";
 import SiteSearchAPIConnector from "@elastic/search-ui-site-search-connector";
 import {
@@ -72,7 +74,7 @@ const config = {
         }
       }
     },
-    disjunctiveFacets: ["acres", "states"],
+    disjunctiveFacets: ["acres", "states", "date_established", "location"],
     facets: {
       world_heritage_site: { type: "value" },
       states: { type: "value", size: 30 },
@@ -94,6 +96,33 @@ const config = {
           { from: 0, to: 100, name: "Nearby" },
           { from: 100, to: 500, name: "A longer drive" },
           { from: 500, name: "Perhaps fly?" }
+        ]
+      },
+      date_established: {
+        type: "range",
+
+        ranges: [
+          {
+            from: moment()
+              .subtract(50, "years")
+              .toISOString(),
+            name: "Within the last 50 years"
+          },
+          {
+            from: moment()
+              .subtract(100, "years")
+              .toISOString(),
+            to: moment()
+              .subtract(50, "years")
+              .toISOString(),
+            name: "50 - 100 years ago"
+          },
+          {
+            to: moment()
+              .subtract(100, "years")
+              .toISOString(),
+            name: "More than 100 years ago"
+          }
         ]
       },
       visitors: {
@@ -182,10 +211,11 @@ export default function App() {
                         view={SingleLinksFacet}
                       />
                       <Facet
-                        field="location"
-                        label="Distance"
+                        field="date_established"
+                        label="Date Established"
                         filterType="any"
                       />
+                      <Facet field="location" label="Distance" />
                       <Facet
                         field="acres"
                         label="Acres"
--- a/packages/react-search-ui-views/src/MultiCheckboxFacet.js
+++ b/packages/react-search-ui-views/src/MultiCheckboxFacet.js
@@ -43,9 +43,12 @@ function MultiCheckboxFacet({
       <div className="sui-multi-checkbox-facet__options-list">
         {options.length < 1 && <div>No matching options</div>}
         {options.map(option => {
-          const checked = !!values.find(value =>
-            deepEqual(option.value, value)
-          );
+          const checked = !!values.find(value => {
+            if (value && value.name && option.value.name === value.name)
+              return true;
+            if (deepEqual(option.value, value)) return true;
+            return false;
+          });
           return (
             <label
               key={`${getFilterValueDisplay(option.value)}`}
--- a/packages/react-search-ui-views/src/SingleSelectFacet.js
+++ b/packages/react-search-ui-views/src/SingleSelectFacet.js
@@ -40,9 +40,16 @@ const setDefaultStyle = {
 function SingleSelectFacet({ className, label, onChange, options, values }) {
   const selectOptions = options.map(toSelectOption);
   const selectedFilterValue = values[0];
-  const selectedOption = selectOptions.find(o =>
-    deepEqual(o.value, selectedFilterValue)
-  );
+  const selectedOption = selectOptions.find(o => {
+    if (
+      selectedFilterValue &&
+      selectedFilterValue.name &&
+      o.value.name === selectedFilterValue.name
+    )
+      return true;
+    if (deepEqual(o.value, selectedFilterValue)) return true;
+    return false;
+  });
 
   return (
     <div className={appendClassName("sui-search-facet sui-facet", className)}>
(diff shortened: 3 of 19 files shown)


packages/react-search-ui-views/src/MultiCheckboxFacet.js
```diff
@@ -43,9 +43,12 @@ function MultiCheckboxFacet({
       <div className="sui-multi-checkbox-facet__options-list">
         {options.length < 1 && <div>No matching options</div>}
         {options.map(option => {
-          const checked = !!values.find(value =>
-            deepEqual(option.value, value)
-          );
+          const checked = !!values.find(value => {
+            if (value && value.name && option.value.name === value.name)
+              return true;
```
constancecchen: This is just for my edification as I get to know the codebase - but do we have Prettier or some linter config that's enforcing column width/this being on a separate line (as opposed to the line below where it's on the same line?)

EDIT: Ah actually I see the `format`/`lint-staged` scripts in the root package.json! Sweet!

packages/react-search-ui-views/src/MultiCheckboxFacet.js
```diff
@@ -43,9 +43,12 @@ function MultiCheckboxFacet({
...
+            if (deepEqual(option.value, value)) return true;
+            return false;
```
constancecchen: I'm a fan of early returns myself, so this feels really nice and easy for me to read personally 👍

packages/react-search-ui-views/src/SingleSelectFacet.js
```diff
@@ -40,9 +40,16 @@ const setDefaultStyle = {
...
   const selectOptions = options.map(toSelectOption);
   const selectedFilterValue = values[0];
-  const selectedOption = selectOptions.find(o =>
-    deepEqual(o.value, selectedFilterValue)
-  );
+  const selectedOption = selectOptions.find(o => {
+    if (
+      selectedFilterValue &&
+      selectedFilterValue.name &&
+      o.value.name === selectedFilterValue.name
```
constancecchen: This is me being a bit nitpicky so feel free to disregard, but I'm personally not a huge fan of single-letter variable names. Something like `option.value.name === selectedFilter.value.name`  would be easier for me to grok personally.

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
```diff
@@ -157,6 +174,37 @@ describe("determining selected option from values", () => {
...
 
+  it("will correctly determine which of the options is selected when using range filters and only the name matches", () => {
+    const wrapper = shallow(
+      <MultiCheckboxFacet
+        {...params}
+        options={rangeOptionsTimeBased}
+        values={[
+          {
+            // The name won't match in this case because even though the user selected and shared "1000 ms ago", the from
+            // is actually different when they reload the page, since it's relative to the current time.
```
constancecchen: I'm not sure I'm reading this comment correctly? I think I get the idea behind it (having had to write date-based tests before) but `the from is actually different from when they reload the page` is throwing me off - is the comment missing a word between `the` and `from`?

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
```diff
@@ -157,6 +174,37 @@ describe("determining selected option from values", () => {
     ).toBe(false);
   });
 
+  it("will correctly determine which of the options is selected when using range filters and only the name matches", () => {
+    const wrapper = shallow(
+      <MultiCheckboxFacet
+        {...params}
+        options={rangeOptionsTimeBased}
+        values={[
```
constancecchen: Just curious - could you pass `values={[rangeOptionsTimeBased[0].value]}` here instead of repeating '1000 ms ago'?

packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
```diff
@@ -3,6 +3,7 @@ import { version } from "../package.json";
 
 import { adaptResponse } from "./responseAdapter";
 import { adaptRequest } from "./requestAdapters";
+import buildResponseAdapterOptionsFromConfig from "./buildResponseAdapterOptionsFromConfig";
```
constancecchen: I think it's going to take me a bit longer to fully read through/understand this new helper, so I'm gonna submit my comments for now and do another pass with the remaining files in a bit here, if that's ok!

My one comment here (feel free to disregard me also if you feel otherwise - I don't feel super strongly about this) is that `buildResponseAdapterOptionsFromConfig` is a bit verbose (although of course very explicit, which is great) - do you think we'd lose anything if we simplified it to, e.g. `buildResponseAdapterOptions` or maybe `buildResponseOptions`?

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
```diff
@@ -0,0 +1,22 @@
+export default function buildResponseAdapterOptionsFromState(config) {
+  const additionalFacetValueFields = Object.entries(config.facets || {}).reduce(
```
constancecchen: Do we need to check for whether `config` is an obj (so that `config.facets` doesn't throw a type error) or is that probably overkill?

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
```diff
@@ -0,0 +1,22 @@
...
+        };
+      }
+
+      return acc;
+    },
+    null
+  );
+
+  return {
+    ...(additionalFacetValueFields && { additionalFacetValueFields })
```
constancecchen: At first I was wondering why you opted to start the reduce above with `null` instead of an empty obj, but this clarifies it. Very clever! 💯

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
```diff
@@ -0,0 +1,22 @@
+export default function buildResponseAdapterOptionsFromState(config) {
```
constancecchen: Naive question here - I'm struggling a bit to understand how this can be expanded to apply to non-geo-specific facets. The name of the function seems to imply that it _generally_ builds options/configs, but the code itself seems pretty limited to geo `unit`/`center` fields.

Do you mind giving me an example of how this might be expanded to cover other facet configs? Or if we don't plan on expanding this down the road, should we maybe consider renaming this something like `buildGeoOptions` or something more specific?

packages/search-ui-app-search-connector/src/__tests__/buildResponseAdapterOptionsFromConfig.test.js
```diff
@@ -0,0 +1,71 @@
...
+          },
+          location: {
+            center: "40.0374748, -76.3046049",
+            unit: "mi"
+          }
+        }
+      });
+    });
+
+    it("should not extract additionalFacetValueFields if there are no geo fields", () => {
```
constancecchen: Just to check about a semi-unrelated corner case:

Will there be any issues / do we want to 'test' what happens if someone passes in a facet obj that has a `center` but no `unit` field, or has `unit` but no `center`? Would elasticsearch still be able to parse that without issue, out of curiosity?

packages/search-ui-app-search-connector/src/__tests__/responseAdapter.test.js
```diff
@@ -8,6 +8,46 @@ describe("adaptResponse", () => {
...
+      additionalFacetValueFields: {
+        location: {
+          center: "73.102, -78.120",
+          unit: "u"
+        }
+      }
+    };
+
+    expect(adaptResponse(response, options).facets.location).toEqual([
+      {
```
constancecchen: I don't feel strongly either way, but I'm seeing most of these objs being defined below as `const`s - do we want to follow that pattern as well and define this as `adaptedLocationResponse` or something similar as well as moving options down as well (maybe as `geoOptions`)? 🤷‍♀

(TBH I'm fine with this as-is, although I'll confess the variable hoisting from the bottom of the file is a bit confusing.)

packages/search-ui/src/__tests__/actions/removeFilter.test.js
```diff
@@ -139,7 +139,46 @@ describe("#removeFilter", () => {
...
+        field: "test",
+        type: "all",
+        values: [
+          { from: "2016-04-11T06:17:43.934Z", name: "A little longer ago" }
+        ]
+      }
+    ]);
+  });
+
+  it("Removes all filters, if no filter value or filter type is specified", () => {
```
constancecchen: nice typo catch/fix here! 💯

packages/search-ui/src/helpers.js
```diff
@@ -23,5 +23,14 @@ export function removeSingleFilterValue(filters, name, value, filterType) {
 }
 
 export function matchFilter(filter1, filter2) {
-  return deepEqual(filter1, filter2);
+  if (
+    filter1 &&
+    filter1.name &&
+    filter2 &&
+    filter2.name &&
+    filter1.name === filter2.name
```
constancecchen: Just for my own context-building - what was the impetus behind adding this extra `.name` checking? Is it fixing an existing issue or is it a performance enhancement?

packages/search-ui/src/__tests__/actions/removeFilter.test.js
```diff
@@ -139,7 +139,46 @@ describe("#removeFilter", () => {
...
+          name: "A really long time ago"
+        },
+        undefined,
+        {
+          initialFilters: [
+            {
+              field: "test",
+              values: [
+                {
+                  from: "1702-07-23T22:17:28.816Z",
```
constancecchen: Just for my own edification - since the new logic is only checking the `name` key, what happens if the `from` key here were totally different from the `from` key on line 147 / if this line were totally removed? The test would still pass, correct?

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
constancecchen: Ah, after coming back to this after having reviewed the removeFilter test down the page - I get this now! It's referring to "the `from` key". I might suggest wrapping the word `from` in backticks to help indicate that it's referring to a var/key name.

examples/sandbox/src/App.js
```diff
@@ -85,6 +87,44 @@ const config = {
           { from: 100001, name: "Large" }
         ]
       },
+      location: {
+        // San Francisco. In the future, make this the user's current position
```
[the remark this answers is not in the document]
JasonStoltz: We do not! Great suggestion though.

packages/react-search-ui-views/src/MultiCheckboxFacet.js
```diff
@@ -43,9 +43,12 @@ function MultiCheckboxFacet({
       <div className="sui-multi-checkbox-facet__options-list">
         {options.length < 1 && <div>No matching options</div>}
         {options.map(option => {
-          const checked = !!values.find(value =>
-            deepEqual(option.value, value)
-          );
+          const checked = !!values.find(value => {
+            if (value && value.name && option.value.name === value.name)
```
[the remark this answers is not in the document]
JasonStoltz: Woah, to be honest I didn't know that was available. That would simplify a lot of our code.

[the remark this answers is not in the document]
JasonStoltz: https://github.com/elastic/search-ui/issues/264

packages/react-search-ui-views/src/SingleSelectFacet.js
JasonStoltz: Cool, done.

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
JasonStoltz: Ugh, yeah, my comments are sometimes not so elegant. I will update this.

packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
JasonStoltz: I appreciate your thoughtfulness. You're giving this way more thought than I did at the time of writing. I frankly just wanted to separate out this code so I could test that logic independently. I will rename it to `buildResponseAdapterOptions`.

packages/search-ui-app-search-connector/src/__tests__/buildResponseAdapterOptionsFromConfig.test.js
JasonStoltz: For simplicity,  I think I am going to just assume them required for now. They are both required field for App Search, which will error with a 4xx if they are missing.

packages/search-ui-app-search-connector/src/__tests__/responseAdapter.test.js
JasonStoltz: I frankly have no opinion on this, so just I'll update it.

JasonStoltz: Except for the hoisting, that's probably a bad practice.

packages/search-ui/src/__tests__/actions/removeFilter.test.js
JasonStoltz: So, correct me if I'm wrong...

Line 147: "1701-07-23T21:15:28.816Z"
Line 157: "1702-07-23T22:17:28.816Z"

They are actually slightly different, is that what you were asking about?

I should just make them really really different so that it is more apparent.

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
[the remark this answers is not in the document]
JasonStoltz: It wouldn't hurt, I can set a default value for config

```
export default function buildResponseAdapterOptionsFromState(config = {}) {
```

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
JasonStoltz: I can rename it to `buildGeoOptions`.

JasonStoltz: Right now it's just a generic function. It takes config, and spits out additional options that need to be passed to the adapter. Down the road, it's possible we may need to pass additional config to the adapter, but as you point out, we don't need to right now.

The code would probably read better if it was just focused on geo options, but I think it would take another step or two of refactoring if we wanted to rename the function to be geo specific, and I frankly don't want to spend the cycles on that right now. Feel free to disagree if you feel strongly.

packages/search-ui/src/__tests__/actions/removeFilter.test.js
constancecchen: OH wow yeah I totally spaced and missed the `1`s that were different from the `2`s! Thats' my bad/eye fatigue or something 🤦‍♀

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
```diff
@@ -181,8 +181,9 @@ describe("determining selected option from values", () => {
         options={rangeOptionsTimeBased}
         values={[
           {
-            // The name won't match in this case because even though the user selected and shared "1000 ms ago", the from
-            // is actually different when they reload the page, since it's relative to the current time.
+            // A time range filter that was applied based on the current time 20 seconds ago, will not have
+            // the same "from" value as a facet that is calculated at the current time. However, we can still
+            // make a match based on the "name" of the values.
```
constancecchen: Love the new comment! 💯

packages/search-ui-app-search-connector/src/__tests__/buildResponseAdapterOptionsFromConfig.test.js
constancecchen: Ahh okay! In that case, what do you think about changing [change line 4 here](https://github.com/elastic/search-ui/pull/260/files#diff-cce8ae85c9de219c2101c466711bcf99R4) to `facetConfig.unit && facetConfig.center` instead of `||`?

packages/search-ui-app-search-connector/src/buildResponseAdapterOptionsFromConfig.js
constancecchen: Nope, I'm good with it! After looking at it a bit more, I also think I see how we could add other branches/if catches! :)

packages/search-ui-app-search-connector/src/__tests__/buildResponseAdapterOptionsFromConfig.test.js
JasonStoltz: Ah, yes, of course.

packages/react-search-ui-views/src/__tests__/MultiCheckboxFacet.test.js
JasonStoltz: Haha, ty.

JasonStoltz:
Thank YOU @constancecchen, for taking the time for thoughtful review. I really like that you took the time to not just glance over the code, but also figure out exactly what is going on and provide very specific, insightful feedback.

Also, I pushed one last fix for your last comment, do you mind re-approving?

PR #266, opened 2019-06-17
Fixed layout issue in Results css

### LINKED ISSUE #265: Layout issue when there are long values in Results
Screenshot 2019-06-17 08 25 18

JasonStoltz:
## Description

Fixed layout issue in Results css when there are long values.

Before:

Screenshot 2019-06-17 09 51 05

After:

Screenshot 2019-06-17 09 46 49

## List of changes

Fixed layout issue in Results css when there are long values.

## Associated Github Issues

Fixes #265

### DIFF
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
@@ -41,6 +41,8 @@
       padding: 2rem 0;
       padding: 32px 0 32px 32px;
       border-left: none;
+      overflow-wrap: break-word;
+      overflow: hidden;
 
       @include block("layout-main-header") {
         display: flex;


yakhinvadim:
Will the wrapping work when there is no "-" and "_" symbols?

JasonStoltz:
@yakhinvadim I don't know.

JasonStoltz:
@yakhinvadim That would probably be a serious edge case.

constancecchen:
@yakhinvadim `overflow-wrap: break-word` is capable of breaking even without the presence of hyphens, so this should still work on extremely long strings. There's some good examples on [MDN's site here](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap).

constancecchen:
@JasonStoltz If the sidebar is never meant to be smaller than its current width, you might also consider changing [this flex line here](https://github.com/JasonStoltz/search-ui/blob/7269dc3935198d282663426da8c960fd918b3589/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss#L35) to `flex: 0 0 30%`, which will set its `flex-shrink` to `0` and means that the column won't compress (which could happen not just for long strings, but, e.g. really wide images/divs, etc.)

PR #270, opened 2019-06-20
Additional options

### LINKED ISSUE #156: Remove or Adapt additionalOptions
I think this should be come two things:

1. "preQuery" hook so a user could adapt queries however they want, and add logic.
2. In "queryConfig" any arbitrary parameters that are provided should be passed through.

JasonStoltz:
## Description

Made `additionalOptions` more useful.

Example:

```js
connector = new AppSearchAPIConnector({
    beforeSearchCall: (existingSearchOptions, next) =>
      next({
        ...existingSearchOptions,
        group: { field: "title" }
      })
  });
```

## List of changes
- Updated hooks to amend, not just append

    Previously the function simply merged in additional options to existing,
    instead, it should allow you to amend the entire object, in case
    there are some you'd like to remove.

- Made hooks work like middleware

    Hook functions now accept a "next" parameter, to allow for adapting
    the response as well, not just the incoming options.

- Separated out `additionalOptions` into 3 distinct hooks:

  `beforeAutocompleteSuggestionsCall`
  `beforeAutocompleteResultsCall`
  `beforeSearchCall`

- Updated docs. Used jsdoc comments and generated markdown for README with https://github.com/jsdoc2md/jsdoc-to-markdown

## Associated Github Issues

Closes #156

### DIFF
--- a/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
+++ b/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
@@ -368,13 +368,14 @@ describe("AppSearchAPIConnector", () => {
     function subject(
       state = {},
       queryConfig = {},
-      beforeAutocompleteResultsCall
+      { beforeAutocompleteResultsCall, beforeAutocompleteSuggestionsCall } = {}
     ) {
       if (!state.searchTerm) state.searchTerm = "searchTerm";
 
       const connector = new AppSearchAPIConnector({
         ...params,
-        beforeAutocompleteResultsCall
+        beforeAutocompleteResultsCall,
+        beforeAutocompleteSuggestionsCall
       });
 
       return connector.onAutocomplete(state, queryConfig);
@@ -562,6 +563,45 @@ describe("AppSearchAPIConnector", () => {
           autocompletedResults: resultState.results,
           autocompletedResultsRequestId: resultState.requestId
         });
+      });
+    });
+
+    describe("beforeAutocompleteResultsCall", () => {
+      it("will use the beforeAutocompleteResultsCall parameter to amend option parameters to the search endpoint call", async () => {
+        await subject(
+          {},
+          { results: {} },
+          {
+            beforeAutocompleteResultsCall: queryOptions => ({
+              ...queryOptions,
+              test: "value"
+            })
+          }
+        );
+        // eslint-disable-next-line no-unused-vars
+        const [_, passedOptions] = getLastSearchCall();
+        expect(passedOptions).toEqual({
+          page: {},
+          test: "value"
+        });
+      });
+    });
+
+    describe("beforeAutocompleteSuggestionsCall", () => {
+      it("will use the beforeAutocompleteSuggestionsCall parameter to amend option parameters to the search endpoint call", async () => {
+        await subject(
+          {},
+          { suggestions: {} },
+          {
+            beforeAutocompleteSuggestionsCall: queryOptions => ({
+              ...queryOptions,
+              test: "value"
+            })
+          }
+        );
+        // eslint-disable-next-line no-unused-vars
+        const [_, passedOptions] = getLastSuggestCall();
+        expect(passedOptions).toEqual({ test: "value" });
       });
     });
 
--- a/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
+++ b/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
@@ -345,17 +345,32 @@ describe("AppSearchAPIConnector", () => {
         current: 2,
         searchTerm: "searchTerm"
       };
-      const beforeSearchCall = (queryOptions, next) =>
-        next({
-          ...queryOptions,
+
+      const queryConfig = {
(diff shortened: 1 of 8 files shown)


packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -340,27 +340,27 @@ describe("AppSearchAPIConnector", () => {
       });
     });
 
-    it("will use the additionalOptions parameter to append additional parameters to the search endpoint call", async () => {
+    it("will use the additionalOptions parameter to amend option parameters to the search endpoint call", async () => {
```
constancecchen: From my understanding of the PR/changes, amending should allow you to remove options from the current obj, correct? Is there any chance we could add/expand these tests to show a value being removed, not just a value being added? That would help my understanding of that functionality personally!

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -340,37 +340,41 @@ describe("AppSearchAPIConnector", () => {
       });
     });
 
-    it("will use the additionalOptions parameter to append additional parameters to the search endpoint call", async () => {
+    it("will use the beforeSearchCall parameter to amend option parameters to the search endpoint call", async () => {
```
constancecchen: From my understanding of the PR/changes, amending should allow you to remove options from the current obj, correct? Is there any chance we could add/expand these tests to show a value being removed, not just a value being added? That would help my understanding of that functionality personally!

JasonStoltz: Yes, great point.

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -561,6 +580,98 @@ describe("AppSearchAPIConnector", () => {
...
+          {
+            beforeAutocompleteSuggestionsCall: (queryOptions, next) =>
+              next({
+                ...queryOptions,
+                test: "value"
+              })
+          }
+        );
+        // eslint-disable-next-line no-unused-vars
+        const [_, passedOptions] = getLastSuggestCall();
```
constancecchen: could we avoid the eslint-disable here by doing `const passedOptions = getLastSuggestCall()[1]` or will that cause a different eslint objection? (no strong feelings either way, just throwing that out there)

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -345,17 +345,32 @@ describe("AppSearchAPIConnector", () => {
...
+      const queryConfig = {
+        sortDirection: "desc",
+        sortField: "name",
+        resultsPerPage: 5
+      };
+
+      const beforeSearchCall = (options, next) => {
+        // Remove sort_direction and sort_field
+        // eslint-disable-next-line no-unused-vars
+        const { sort, ...rest } = options;
```
constancecchen: just curious: would a dev be able to `delete options.sort` instead? or would they not want to do that since it's mutating the original `options` obj?

constancecchen:
I super love the new jsdoc comment blocks, they're very helpful & explicit! Also, thanks so much for the atomic/logical commit progression - it definitely made this easier to follow along 😄 

One thing I'm not sure I've 100% grokked with the 3 new `beforeX/Y/Z` hooks:

1. I assume that these are all optional and don't need to be included in the config obj if you don't wish to change anything. Is that assumption correct?

2. If so - for my edification, where's the fallback for these hooks being declared? I'm guessing in the [constructors](https://github.com/elastic/search-ui/pull/270/files#diff-b481ce00297f52a186b143f675c3f0d0R52) - does that assumption sound correct?

3. One thing I haven't quite grasped - where's the fallback for the `next` function/callback being declared? If it's missing / not declared, wouldn't that be a type/definition error trying to run undefined as a function?

Thanks so much for answering my newb questions! 🙏

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
JasonStoltz: @constancecchen I *think* they could do that without any adverse side effects.

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
JasonStoltz: Yes, I think that would be better.

JasonStoltz:
Hey @constancecchen,

1) Correct
2) Also correct
3) `next` is a parameter that the Connector passes to the callback functions. 
https://github.com/JasonStoltz/search-ui/blob/additionalOptions/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js#L114. So they won't be `undefined`, that's completely under our control. Does that makes sense?

constancecchen:
Ohhhh gotcha gotcha! My confusion was thinking that `next` was being defined/passed by the user/developer. That clears it up totally - thanks!

Issue #256, opened 2019-06-10
Schema for `Result` objects is too strict


JasonStoltz:
Currently does not allow for nested objects.

May also not allow for other valid types, like Boolean, etc.

conor909:
Does anyone know if this has been released?   I can't seem to use nested objects in the responses in react-search-ui

`v1.2.0`

JasonStoltz:
@conor909 It was, can you be more specific? What are you using this with, App Search or Elasticsearch? What does the object look like.

JasonStoltz:
And yes, this was released in 0.11

conor909:
Im using the example project in `examples/elasticsearch` and I'm testing it with the json file in `examples/elasticsearch/data/national-es-bulk-index.json` which works great.

But if I change the structure of the JSON to have a nested object, then it breaks because it can't find the path to the `raw` property.

JasonStoltz:
@conor909 This PR loosened the schema for the that Results type. However, the out of the box `Results` view component *does* expect the data to be formatted in a particular way. The view unfortunately does not currently communicate that properly in it's prop types: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui-views/src/Result.js#L102

You would need to provide a custom view for your application.

Here's a description of custom view: https://github.com/elastic/search-ui/blob/master/ADVANCED.md#component-views-and-html.

And an example you can use for reference: https://codesandbox.io/s/video-games-tutorial-with-images-5wsqs.

conor909:
@JasonStoltz ah I see, thanks for that clear info!

PR #271, opened 2019-06-21
Loosened the "Result" object schema

JasonStoltz:
## Description

In our type definition for a "Result", we required it to have
a raw and snippet key:

```
{
  raw: "",
  snippet: ""
}
```

However, while this is the most common scenario, it's actually
technically possible to get other stuff back in a "Result".
For instance, when using the "group" feature in the App Search API,
you can get back a "_group" key in results. Or if you are using
this project with Elasticsearch, you might get a nested object back.
This schema was too arbitrarily strict. It's better to accept
anything here.

## List of changes
- Expanded results shown in the Sandbox application, just for a better demo and also for testing.
- Updated BOTH `Result.js` types, which are duplicated in two separate packages currently, but we should definitely find a way to consolidate in the future.
- Changed the `Result.js` view to filter out arbitrary `Result` values, since we've now loosened the schema. It only attempts to render things that look like a `FieldValueWrapper` (i.e., they have a `raw` and `snippet` value, which is how App Search returns field values).

## Associated Github Issues

Fixes: #256

### DIFF
--- a/packages/react-search-ui-views/src/Result.js
+++ b/packages/react-search-ui-views/src/Result.js
@@ -2,6 +2,7 @@ import PropTypes from "prop-types";
 import React from "react";
 
 import { appendClassName } from "./view-helpers";
+import { isFieldValueWrapper } from "./types/FieldValueWrapper";
 
 function getFieldType(result, field, type) {
   if (result[field]) return result[field][type];
@@ -36,6 +37,14 @@ function getEscapedField(result, field) {
 
 function getEscapedFields(result) {
   return Object.keys(result).reduce((acc, field) => {
+    // Because if it is an arbitrary value from the response, we may not properly
+    // handle it, so we filter out arbitrary values here.
+    //
+    // I.e.,
+    // Arbitrary value: "_metaField: '1939191'"
+    // vs.
+    // FieldValueWrapper: "_metaField: {raw: '1939191'}"
+    if (!isFieldValueWrapper(result[field])) return acc;
     return { ...acc, [field]: getEscapedField(result, field) };
   }, {});
 }
@@ -67,12 +76,12 @@ function Result({ className, result, onClickLink, titleField, urlField }) {
       </div>
       <div className="sui-result__body">
         <ul className="sui-result__details">
-          {Object.keys(fields).map(key => (
-            <li key={key}>
-              <span className="sui-result__key">{key}</span>{" "}
+          {Object.entries(fields).map(([fieldName, fieldValue]) => (
+            <li key={fieldName}>
+              <span className="sui-result__key">{fieldName}</span>{" "}
               <span
                 className="sui-result__value"
-                dangerouslySetInnerHTML={{ __html: fields[key] }}
+                dangerouslySetInnerHTML={{ __html: fieldValue }}
               />
             </li>
           ))}
--- a/packages/react-search-ui-views/src/Result.js
+++ b/packages/react-search-ui-views/src/Result.js
@@ -37,8 +37,8 @@ function getEscapedField(result, field) {
 
 function getEscapedFields(result) {
   return Object.keys(result).reduce((acc, field) => {
-    // Because if it is an arbitrary value from the response, we may not properly
-    // handle it, so we filter out arbitrary values here.
+    // If we receive an arbitrary value from the response, we may not properly
+    // handle it, so we should filter out arbitrary values here.
     //
     // I.e.,
     // Arbitrary value: "_metaField: '1939191'"
--- a/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
+++ b/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
@@ -1,5 +1,41 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`filters out arbitrary fields from results, and does not render them 1`] = `
+<li
+  className="sui-result"
+>
+  <div
+    className="sui-result__header"
+  />
+  <div
+    className="sui-result__body"
+  >
+    <ul
+      className="sui-result__details"
+    >
+      <li
+        key="field"
+      >
+        <span
+          className="sui-result__key"
+        >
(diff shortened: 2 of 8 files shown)


packages/react-search-ui-views/src/Result.js
```diff
@@ -36,6 +37,14 @@ function getEscapedField(result, field) {
 
 function getEscapedFields(result) {
   return Object.keys(result).reduce((acc, field) => {
+    // Because if it is an arbitrary value from the response, we may not properly
+    // handle it, so we filter out arbitrary values here.
+    //
+    // I.e.,
+    // Arbitrary value: "_metaField: '1939191'"
+    // vs.
+    // FieldValueWrapper: "_metaField: {raw: '1939191'}"
```
constancecchen: This example personally really helped me - really appreciate the comment :)

packages/react-search-ui-views/src/Result.js
```diff
@@ -67,12 +76,12 @@ function Result({ className, result, onClickLink, titleField, urlField }) {
...
-          {Object.keys(fields).map(key => (
-            <li key={key}>
-              <span className="sui-result__key">{key}</span>{" "}
+          {Object.entries(fields).map(([fieldName, fieldValue]) => (
+            <li key={fieldName}>
+              <span className="sui-result__key">{fieldName}</span>{" "}
               <span
                 className="sui-result__value"
-                dangerouslySetInnerHTML={{ __html: fields[key] }}
+                dangerouslySetInnerHTML={{ __html: fieldValue }}
```
constancecchen: this reads a little more nicely to me as well 💯

packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
```diff
@@ -1,5 +1,41 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`filters out arbitrary fields from results, and does not render them 1`] = `
```
constancecchen: Just curious / for my own edification (I haven't used Jest's snapshots very much in the past, although now I'm thinking I should!) - why have the string within `exports['foo']` be a backtick/interpolated string? I get why the JSX itself is (for multiple lines), but less so the array key. Is it just a Jest standard?

packages/react-search-ui-views/src/types/Result.js
```diff
@@ -2,6 +2,15 @@ import PropTypes from "prop-types";
 
 import FieldValueWrapper from "./FieldValueWrapper";
 
-// An object where keys are the field names, and values are the values
-// corresponding to that
-export default PropTypes.objectOf(FieldValueWrapper);
+// Typically an object where keys are the field names, and values are field values.
+// Also could be literally any other arbitrary value depending on the particular Search API response.
+// Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
+// is typically better to work with FieldValueWrapper values. So we encourage FieldValueWrapper, but we accept
```
constancecchen: v minor copy thing, I'd suggest replacing `typically` here with `usually` or `generally`, since we've already used `typically` at the start of the comment.

I'd also lose the `So` as well since we already used 'so' just one sentence ago, and the sentence reads fine without it.

```suggestion
// is usually better to work with FieldValueWrapper values. We encourage FieldValueWrapper, but we accept
```

packages/react-search-ui/src/types/Result.js
```diff
@@ -2,5 +2,15 @@ import PropTypes from "prop-types";
 
 import FieldValueWrapper from "./FieldValueWrapper";
 
-// An object where keys are the field names, and values are field values
-export default PropTypes.objectOf(FieldValueWrapper);
+// Typically an object where keys are the field names, and values are field values.
+// Also could be literally any other arbitrary value depending on the particular Search API response.
+// Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
+// is typically better to work with FieldValueWrapper values. So we encourage FieldValueWrapper, but we accept
```
constancecchen: Same copy comment as above here

```suggestion
// is usually better to work with FieldValueWrapper values. We encourage FieldValueWrapper, but we accept
```

packages/react-search-ui/src/types/FieldValueWrapper.js
```diff
@@ -3,9 +3,17 @@ import PropTypes from "prop-types";
 import FieldValue from "./FieldValue";
 
 export default PropTypes.shape({
-  // A raw field value, like 'I am a raw result', or 2, or true
+  // A raw field value, like 'I am a raw result', or 2, or true. Raw values may
+  // or may not be html escaped, so *always* sanitize a raw value before rendering
+  // it on a page as html.
```
constancecchen: Super awesome comment/context additions! 💯

packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
JasonStoltz: @constancecchen These are actually auto-generated by Jest, so not sure why they do that. Jest snapshots are great *sometimes*. There are mixed opinions, for sure.

packages/react-search-ui-views/src/Result.js
```diff
@@ -36,6 +37,14 @@ function getEscapedField(result, field) {
 
 function getEscapedFields(result) {
   return Object.keys(result).reduce((acc, field) => {
+    // Because if it is an arbitrary value from the response, we may not properly
+    // handle it, so we filter out arbitrary values here.
```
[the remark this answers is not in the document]
JasonStoltz: Updated, ty.

JasonStoltz:
@constancecchen Great, ty. Ready for another look.

packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
constancecchen: Ohhh gotcha gotcha, super interesting to know!

PR #272, opened 2019-06-21
Gliiiiiiiiitch

JasonStoltz:
## Description

Gliiiiiiiiitch

## List of changes

- Gliiiiiiiiitch

## Associated Github Issues

Fixes #Gliiiiiiiiitch

Screenshot 2019-06-21 16 12 04

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 ## About Search UI :rocket:
 
-A **[React](https://reactjs.org)**  library for [**Elastic App Search**](https://www.elastic.co/cloud/app-search-service) that allows you to quickly implement search experiences without re-inventing the wheel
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -32,7 +32,9 @@ A **[React](https://reactjs.org)**  library for [**Elastic App Search**](https:/
 
 ### Live Demo
 
-Checkout the [live demo of Search UI](https://search-ui-stable.netlify.com).
+Checkout the [live demo of Search UI](https://search-ui-parks-example.glitch.me).
```
constancecchen: My 2c for what it's worth - I think I semi prefer the old netlify link just because it doesn't have the loading/spinup time that Glitch does. That being said I'm all for keeping/adding the new "Remix on Glitch" link below!

JasonStoltz: @constancecchen I think I agree

Issue #163, opened 2019-04-10
Allow a user to hide the button on SearchBox


JasonStoltz:
```jsx
<SearchBox inputView={ ( { getTextInputProps, getSubmitButtonProps } ) => (
    <input {...getTextInputProps({
       placeholder: "Search your documents",
       // You can inject or override whatever attributes you want here
     })} />
    <input {...getSubmitButtonProps({
       className: "some custom class"
       // You can inject or override whatever attributes you want here
     })} />
) } />
```

So I would replace the `inputProps` with an `inputView` function. You could then arrange the text input and submit button however you'd like. You could choose to omit the button completely in this case. Or, you could choose to apply a class, or whatever you'd like. You could inject an icon if you needed to do that... this would give you the maximum amount of flexibility I believe.

The only thing you would need to do is make sure when you are applying attributes to either element, that you are applying them through `getTextInputProps ` or `getSubmitButtonProps`. Those functions will make sure any props are applied that MUST be applied for them to work correctly, but also give you the opportunity to mix in your own.

katydecorah:
I was thinking that this could be implement as an object, `buttonProps`, that gets passed to the button much like `inputProps`. This way the button can be customize with class names. In my use case, I would hide the button, but passing props would make the component highly customizable.

JasonStoltz:
@katydecorah The approach I had in mind looks something like this:

```jsx
<SearchBox inputView={ ( { getTextInputProps, getSubmitButtonProps } ) => (
    <input {...getTextInputProps({
       placeholder: "Search your documents",
       // You can inject or override whatever attributes you want here
     })} />
    <input {...getSubmitButtonProps({
       className: "some custom class"
       // You can inject or override whatever attributes you want here
     })} />
) } />
```

So I would replace the `inputProps` with an `inputView` function. You could then arrange the text input and submit button however you'd like. You could choose to omit the button completely in this case. Or, you could choose to apply a class, or whatever you'd like. You could inject an icon if you needed to do that... this would give you the maximum amount of flexibility I believe.

The only thing you would need to do is make sure when you are apply attributes to either element, that you are applying them through `getInputProps` or `getSubmitButtonProps`. Those functions will make sure any props are applied that MUST be applied for them to work correctly, but also give you the opportunity to mix in your own.

Does that makes sense?

@scottybollinger and @zumwalt I think this would solve a lot of your issues with this component as well.

katydecorah:
> Does that makes sense?

[...]

PR #273, opened 2019-06-24
Added a inputView option to SearchBox

JasonStoltz:
## Description

Adding this allows for much simpler, targeted view customization.
Instead of having to deal with all of the Downshift plumbing and logic,
this lets you very easily target just the html.

For example, if you just want to hide the search button, or inject an icon, etc. This makes it much easier. CC @scottybollinger @zumwalt, this is something you had dealt with before.

```jsx
                    <SearchBox
                      inputView={({
                        getAutocomplete,
                        getButtonProps,
                        getInputProps
                      }) => (
                        <>
                          <div className="sui-search-box__wrapper">
                            <input
                              {...getInputProps({ placeholder: "I am thor" })}
                            />
                            {getAutocomplete()}
                          </div>
                          <input {...getButtonProps()} />
                        </>
                      )}
```

See the ADVANCED.md updates for some more examples.

## List of changes

- Added a inputView option to SearchBox

## Associated Github Issues

#163

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -314,20 +314,37 @@ customizations, like simply hiding the search button, this is often overkill.
 ```
 
 You can also just customize the input section of the search box. Useful for things
-like hiding the submit button:
+like hiding the submit button or rearranging dom structure:
 
 ```jsx
 <SearchBox
-  inputView={({ getAutocomplete, getInputProps }) => (
+  inputView={({ getAutocomplete, getInputProps, getButtonProps }) => (
     <>
       <div className="sui-search-box__wrapper">
-        <input {...getInputProps()} />
+        <input
+          {...getInputProps({
+            placeholder: "I am a custom placeholder"
+          })}
+        />
         {getAutocomplete()}
+        <input
+          {...getButtonProps({
+            "data-custom-attr": "some value"
+          })}
+        />
       </div>
     </>
   )}
 />
 ```
+
+Note that `getInputProps` and `getButtonProps` are meant to be spread over their corresponding
+UI elements. This lets you arrange elements however you'd like in the DOM. It also lets you pass
+additional properties. You should pass properties through these functions, rather directly on elements,
+in order to not override base values. For instance, adding a `className` through these functions will
+assure that the className is only appended, not overriding base class values on those values.
+
+`getAutocomplete` is used to determine where the autocomplete dropdown will be shown.
 
 Or you can also just customize the autocomplete dropdown:
 
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -338,11 +338,12 @@ like hiding the submit button or rearranging dom structure:
 />
 ```
 
-Note that `getInputProps` and `getButtonProps` are meant to be spread over their corresponding
-UI elements. This lets you arrange elements however you'd like in the DOM. It also lets you pass
-additional properties. You should pass properties through these functions, rather directly on elements,
-in order to not override base values. For instance, adding a `className` through these functions will
-assure that the className is only appended, not overriding base class values on those values.
+Note that `getInputProps` and `getButtonProps` are
+[prop getters](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters).
+They are meant return a props object to spread over their corresponding UI elements. This lets you arrange
+elements however you'd like in the DOM. It also lets you pass additional properties. You should pass properties
+through these functions, rather directly on elements, in order to not override base values. For instance,
+adding a `className` through these functions will assure that the className is only appended, not overriding base class values on those values.
 
 `getAutocomplete` is used to determine where the autocomplete dropdown will be shown.
 
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -336,7 +336,6 @@ Or you can also just customize the autocomplete dropdown:
   autocompleteView={({ autocompletedResults, getItemProps }) => (
     <div className="sui-search-box__autocomplete-container">
       {autocompletedResults.map((result, i) => (
-        // eslint-disable-next-line react/jsx-key
         <div
           {...getItemProps({
             key: result.id.raw,
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -286,7 +286,7 @@ import { SearchBox } from "@elastic/react-search-ui";
 
 ...
 
-<SearchBox inputProps={{ placeholder: "custom placeholder" }}/>
+<SearchBox />
 ```
 
 ### Configuring search queries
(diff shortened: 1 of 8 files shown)


packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -53,25 +56,35 @@ function SearchBox(props) {
                 appendClassName("sui-search-box", className) + autocompleteClass
               }
             >
-              <div className="sui-search-box__wrapper">
-                <input
-                  {...getInputProps({
+              <InputView
+                getInputProps={additionalProps =>
```
constancecchen: I don't think I'm understanding where `additionalProps` are coming from or how devs should pass this in in a final usage situation. Is there any chance we could write an extra test that shows a custom `inputView` function with additional props for both the search input and the button input?

packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -53,25 +56,35 @@ function SearchBox(props) {
...
+                })}
+                getAutocomplete={() => {
+                  if (
+                    useAutocomplete &&
+                    isOpen &&
+                    allAutocompletedItemsCount > 0
+                  ) {
+                    return <AutocompleteView {...props} {...downshiftProps} />;
+                  } else {
+                    return null;
```
constancecchen: Very minor code style suggestion - but you might get fewer lines/a smoother flow by turning this into a ternary: 

```jsx
getAutocomplete={() => {
  const showAutocomplete =
    (useAutocomplete && isOpen && allAutocompletedItemsCount > 0);
  return showAutocomplete
    ? <AutocompleteView {...props} {...downshiftProps} />
    : null;
}}
```

packages/react-search-ui-views/src/SearchBox.js
constancecchen: Ah, duh, I see an example in the Storybook file below. I'd still really like to see a test written for the new functionality possible, but totally get if not or if it's better suited for an integration test.

I'm also a little curious about the purpose of the markdown file examples vs the storybook examples - would it make sense to combine them at one point so all examples are in one location? Or are they specifically split up for a reason?

ADVANCED.md
```diff
@@ -295,6 +295,62 @@ The input from `SearchBox` will be used to trigger a new search query. That quer
...
+```
+
+Or you can also just customize the autocomplete dropdown:
+
+```jsx
+<SearchBox
+  autocompleteView={({ autocompletedResults, getItemProps }) => (
+    <div className="sui-search-box__autocomplete-container">
+      {autocompletedResults.map((result, i) => (
+        // eslint-disable-next-line react/jsx-key
```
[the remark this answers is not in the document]
JasonStoltz: Yes, will remove it.

packages/react-search-ui-views/src/SearchInput.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+function SearchInput({ getAutocomplete, getButtonProps, getInputProps }) {
```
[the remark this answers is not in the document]
JasonStoltz: I don't have a particular reason for doing it one way or the other.

packages/react-search-ui-views/src/SearchBox.js
JasonStoltz: Sometimes the explicit if / else is just easier to read. I don't mind changing it though.

constancecchen: I'm good either way!

packages/react-search-ui-views/src/SearchInput.js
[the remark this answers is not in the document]
constancecchen: Cool, no worries! I'll keep this format in mind if I ever contribute/write a component! 🤞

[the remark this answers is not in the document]
JasonStoltz: @constancecchen I am lax on code style. I have probably used both styles for functional components.

packages/react-search-ui-views/src/SearchBox.js
JasonStoltz: The markdown files are the documentation, meant for people using the project. The storybook is more for us, being able to test and styles components in various states in isolation. I would very much like to consolidate this. I think, ideally, that component reference section could just be replaced with storybook.

JasonStoltz: @constancecchen I didn't realize that we don't have a test for this. It was in severe need. I added tests that cover both `getInputProps` and `getButtonProps`. I also did some reworking so they preserve css class names, which would have been a big issue. Thank you for your keen eyes.

I unfortunately could not find a way to test `getAutocomplete`, but I need to move on from this PR. The smoke tests in storybook and the sandbox will have to be enough for now.

I also added notes about this pattern to the README. I'd like to expand more views, in the future, to use this pattern: https://github.com/downshift-js/downshift#prop-getters

JasonStoltz:
@constancecchen Ready for another look!

packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -57,23 +57,31 @@ function SearchBox(props) {
               }
             >
               <InputView
-                getInputProps={additionalProps =>
-                  getInputProps({
+                getInputProps={additionalProps => {
+                  const { className, ...rest } = additionalProps || {};
```
constancecchen: Awesome catch here! 👏

packages/react-search-ui-views/src/__tests__/SearchBox.test.js
```diff
@@ -60,7 +60,95 @@ it("applies className from inputProps to input element", () => {
...
+      .shallow()
+      .find("input")
+      .at(0);
+
+    button = wrapper
+      .dive("Downshift")
+      .find("inputView")
+      .shallow()
+      .find("input")
+      .at(1);
```
constancecchen: Good ol' Jest chaining + prettier autolinting :) This is a really clever way of DRYing out the various child diving/shallowing, huge shout out! 💯 

Optional suggestion / feel free to disregard this/ship as-is (like you mentioned, probably good to move on at this point), but if you wanted to be even DRYer and possibly sidestep prettier's linting you could try something like this:

```js
inputView = wrapper
  .dive("Downshift")
  .find("inputView")
  .shallow(); // I think Prettier still multi-line this, oh well
input = inputView.find('input').at(0);
button = inputView.find('button').at(1);
```
and then down on line 113 you can reuse inputView again:
```js
expect(inputView.find('.some_custom_wrapper_class')).toHaveLength(1);
```

packages/react-search-ui-views/src/SearchBox.js
constancecchen: All good, this is an amazing amount of extra tests/documentation!! Thanks so much for going above and beyond! 😁

ADVANCED.md
```diff
@@ -314,21 +314,38 @@ customizations, like simply hiding the search button, this is often overkill.
 ```
 
 You can also just customize the input section of the search box. Useful for things
-like hiding the submit button:
+like hiding the submit button or rearranging dom structure:
```
constancecchen: Super minor grammar nit: DOM should probably be capitalized here / for consistency (it's capitalized elsewhere on the page)

JasonStoltz: Good point! Sorry, hit the merge button.

packages/react-search-ui-views/src/__tests__/SearchBox.test.js
JasonStoltz: Good point! Sorry, hit the merge button.

PR #291, opened 2019-06-26
Fixed issue with state not updating

### LINKED ISSUE #279: WithSearch under conditionally rendered components return previous instead of current state
# Steps to reproduce

Code: https://codesandbox.io/s/search-ui-national-parks-example-3qb66
Recording: https://zappy.zapier.com/F2516A6C-523A-4F14-B6CE-C94DF9A26859.mp4

1. Open https://3qb66.csb.dev/?q=Rocky&size=n_20_n
2. Notice how the conditionally rendered components do not list the results, while the unconditional one does.
3. Start changing the search term to _Mountain_
4. Notice how while you type the component both conditional components get updated to render the "old" results.
5. Hit enter to kick off the new search.
6. Notice how the unconditional and the component rendered conditionally on `wasSearch` do render the new results, but the other conditional component (which while `isLoading` doesn't render) still shows the old ones.
7. GOTO 3) 😉 

# Theory

I think this is due to how `WithSearch` used within these components/containers subscribes to changes when it gets constructed:

https://github.com/elastic/search-ui/blob/ba689c64f28ae65b84064be903367687e7547c7f/packages/react-search-ui/src/withSearch.js#L52

If the component gets constructed it somehow gets the previous instead of the current state.

JasonStoltz:
## Description

This fixes an issue described in #279, where if you had components
nested under a `WithSearch`, that used state from `WithSearch` to
conditionally render, they would be one state update behind.

This is because the conditional components are initialized after
a state update on the higher level `WithSearch`, who's update was
triggered because they are subscribed to the driver's state.

When the nested component's constructor runs, they grab the most
recent state from the driver. However, the way the driver was set up,
subscriptions were executed BEFORE the driver's own state was updated.
Therefore, the nested component's state was being derived from the
outdated driver state.

## List of changes

- Changed driver to update own state before notifying subscriptions

## Associated Github Issues

Fixes #279

### DIFF
--- a/packages/search-ui/src/SearchDriver.js
+++ b/packages/search-ui/src/SearchDriver.js
@@ -272,8 +272,8 @@ export default class SearchDriver {
     const state = { ...this.state, ...newState };
     // eslint-disable-next-line no-console
     if (this.debug) console.log("State Update", newState, state);
-    this.subscriptions.forEach(subscription => subscription(state));
     this.state = state;
+    this.subscriptions.forEach(subscription => subscription(state));
   }
 
   /**
--- a/packages/search-ui/src/__tests__/SearchDriver.test.js
+++ b/packages/search-ui/src/__tests__/SearchDriver.test.js
@@ -279,6 +279,21 @@ describe("subscribeToStateChanges", () => {
     expect(called1).toBe(true);
     expect(called2).toBe(true);
   });
+
+  it("will update own state before notifying subscribers", () => {
+    const { driver } = setupDriver();
+    let searchTermFromDriver, searchTermFromSubscription, called;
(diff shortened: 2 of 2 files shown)


FokkeZB:
Awesome @JasonStoltz! Any chance we could have a patch release soon? I have a workaround in place (one big WithSearch passing everything down, skipping containers) but would like to drop that ASAP.

JasonStoltz:
@FokkeZB I'm going to release a new, backwards compatible minor version today.

JasonStoltz:
@FokkeZB https://github.com/elastic/search-ui/releases/tag/v0.11.0. Let me know if you have any additional issues.

Issue #293, opened 2019-06-27
I don't think our stylesheet should rely on a hosted font


JasonStoltz:
Currently, an network request is made for https://fonts.googleapis.com if you include the search-ui CSS in a project.

I think this should be inlined if it needs to be a dependency.

https://github.com/elastic/search-ui/blob/121616c4d6647a9f4bc059869d620c083e0b90eb/packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss#L9

Any thoughts on this?

yakhinvadim:
I think we shouldn't import fonts inside the library. That will allow users to import fonts the way they need it. For example, they might want to import them from their own CDN, or they might want to use display-font property inside their font-face declaration for performance reasons - there was no option to do that in Google Fonts until only a month ago. There could be other reasons to do that.

That would not take much from users, they'll just need to insert the import line anywhere in their own code. But it will give flexibility and performance.

Even the Material UI library, which very highly depends on Roboto asks users to install the font themselves: https://material-ui.com/getting-started/installation/#roboto-font

constancecchen:
Yep, I think [Google Font's new `font-display` support](https://www.zachleat.com/web/google-fonts-display/) alone is a good reason to have users self-link. Performance++ :)

My suggestion/personal preference would be having the default font for Search UI just be [system fonts](https://css-tricks.com/snippets/css/system-font-stack/) (e.g. San Francisco for Mac, Segoe for Windows). Simple, good-looking, and consistent for many users.

zumwalt:
The system font stack we usually use looks something like this: `font-family: system, -apple-system, ".SFNSText-Regular", "San Francisco", "Roboto", "Segoe UI", "Helvetica Neue", "Lucida Grande", sans-serif;`

PR #296, opened 2019-06-28
Removed Roboto font

JasonStoltz:
## Description

Please compare the following two sites.

Without Roboto (new): https://search-ui-canary.netlify.com
With Roboto (current): https://search-ui-stable.netlify.com

The difference is very subtle, and I think this will be fine.

## List of changes

- Removed Roboto font and replaced with system styles suggested by @zumwalt 

## Associated Github Issues

Fixes: https://github.com/elastic/search-ui/issues/293

### DIFF
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
@@ -1,12 +1,12 @@
 html {
   box-sizing: border-box;
 }
-*, *:before, *:after {
+*,
+*:before,
+*:after {
   box-sizing: inherit;
 }
 
-
-@import url("https://fonts.googleapis.com/css?family=Roboto:400,700");
 @import "modules/variables";
 @import "modules/mixins";
 @import "layouts/layout";
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
@@ -30,7 +30,8 @@ $fieldHoverColor: #f6f6f6 !default;
 $resultKeyTextColor: #777777 !default;
 
 // Font Variables
-$fontFamily: 'Roboto', sans-serif !default;
+$fontFamily: system, -apple-system, ".SFNSText-Regular", "San Francisco",
+  "Roboto", "Segoe UI", "Helvetica Neue", "Lucida Grande", sans-serif !default;
 $fontWeightNormal: 400 !default;
 $fontWeightBold: 700 !default;
 $lineHeight: 1.5 !default;
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
@@ -30,8 +30,8 @@ $fieldHoverColor: #f6f6f6 !default;
 $resultKeyTextColor: #777777 !default;
 
 // Font Variables
-$fontFamily: system, -apple-system, ".SFNSText-Regular", "San Francisco",
-  "Roboto", "Segoe UI", "Helvetica Neue", "Lucida Grande", sans-serif !default;
+$fontFamily: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica,
+  Arial, sans-serif !default;
 $fontWeightNormal: 400 !default;
 $fontWeightBold: 700 !default;
 $lineHeight: 1.5 !default;


yakhinvadim:
I don't know the code that well yet, is there a way for users to use their own font in search-ui?

packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
```diff
@@ -1,12 +1,12 @@
 html {
   box-sizing: border-box;
 }
-*, *:before, *:after {
+*,
+*:before,
+*:after {
```
constancecchen: Since we're already here / linting stuff - pseudo classes should [prefer having two colons](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Syntax) instead of one to separate pseudo elements vs. pseudo classes (e.g. `:hover`). It might be nice to fix this to `::before` and `::after` 😄

packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
```diff
@@ -30,7 +30,8 @@ $fieldHoverColor: #f6f6f6 !default;
 $resultKeyTextColor: #777777 !default;
 
 // Font Variables
-$fontFamily: 'Roboto', sans-serif !default;
+$fontFamily: system, -apple-system, ".SFNSText-Regular", "San Francisco",
+  "Roboto", "Segoe UI", "Helvetica Neue", "Lucida Grande", sans-serif !default;
```
constancecchen: I think I'm actually going to push back a bit on Lucida Grande and Helvetica Neue here. Those are generally used by older Mac systems that I don't think it makes a ton of sense to support/target if we're not supporting non-evergreen browsers in any case.

My suggestion would be to take a leaf from what GitHub/Medium/WordPress are doing for their system fonts snippets (I'm assuming that they have more resources than us to do a bunch of OS testing!), which is shorter and closer to:

```
-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
```

packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
yakhinvadim: I agree. Although I'd prefer having something like "stylelint --fix" doing this for us :smile:. I'll keep it in mind for the future.

constancecchen: Yeah! Strong agree to having stylelint+prettier doing this kind of thing automatically for us ✨

packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
constancecchen: BTW - we definitely need to avoid just the `system` font-family. We want either `-apple-system` or `system-ui` (usually just one is fine), but not `system`. [Medium actually ran into this a while ago](https://medium.design/system-shock-6b1dc6d6596f) where `system` on Windows is an old and incredibly retro/pixelated font lol.

packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
JasonStoltz: I'm not going to fix that in this PR, but I do like your suggestion.

packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
JasonStoltz: @zumwalt Do you have an opinion here?

Constance, what is the full font-family value you are proposing.

I don't hav an opinion, I will put whatever you want in here.

JasonStoltz:
> I'm happy with the changes, but I have one question: I don't know the code that well yet, is there a way for users to use their own font in search-ui?
> I think some users might not care too much about font loading performance and would prefer having their custom fonts instead of system fonts.

@yakhinvadim Nope, the only way to do anything like that is to provide your own stylesheet currently.

packages/react-search-ui-views/src/styles/themes/reference-ui/_base.scss
constancecchen: Cool, no worries. :)

packages/react-search-ui-views/src/styles/themes/reference-ui/modules/_variables.scss
constancecchen:
```
-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
```
Would be my suggestion.

zumwalt: I agree with @constancecchen.

yakhinvadim:
> > I'm happy with the changes, but I have one question: I don't know the code that well yet, is there a way for users to use their own font in search-ui?
> > I think some users might not care too much about font loading performance and would prefer having their custom fonts instead of system fonts.
> 
> @yakhinvadim Nope, the only way to do anything like that is to provide your own stylesheet currently.

This is a way I'm ok with for now 🙂

PR #300, opened 2019-07-01
Fix broken relative links in ADVANCED.md

constancecchen:
## Description
I noticed a few broken links in the ADVANCED docs while working on my onboarding project, and thought I'd open up a quick PR to fix them here.

Note: If you'd rather keep these as relative links instead of absolute, and only fix the pathing instead, I can do this - just let me know!

## List of changes
- Fix a few broken relative links in ADVANCED.md

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -218,14 +218,14 @@ Request state can be set by:
 - The `initialState` option.
 - The URL query string, if `trackUrlState` is enabled.
 
-| option                                            | type                                                                                                           | required? | source                                 |
-| ------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- | --------- | -------------------------------------- |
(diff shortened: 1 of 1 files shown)


JasonStoltz:
@constancecchen I prefer relative links, because currently, if someone is viewing an older version of the docs / this repository, I'd like them to link in a way that keeps them in that version, and does not send them to master.

constancecchen:
Good call! I've pushed up a commit that changes docs links to relative (which includes some URLs that were previously absolute that I've now changed to relative).

As a heads up, there's still one absolute link remaining at the top of the page - `@elastic/search-ui`, which links to `https://github.com/elastic/search-ui/tree/master/packages/search-ui` - but I think that one has to stay absolute due to the `/tree/` path. Thoughts?

PR #301, opened 2019-07-02
Fixed a CSS bug

JasonStoltz:
## Description

After removing the Roboto font, there was an issue with the
ResultPerPage dropdown, where it was too narrow for the font and
was being squashed down to an ellipses.

Additionally, there was an issue where the box shadow of the
dropdown was being cut off due to a change made in #266. I moved
the styles from that change to be more targeted to fix this.

Before:

css-iussues

After:

css-fix

### DIFF
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_select.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_select.scss
@@ -4,7 +4,7 @@
   font-size: 0.875rem;
   margin-top: $sizeS;
   width: 100%;
-  min-width: 60px;
+  min-width: 65px;
 
   @include modifier('inline') {
     margin-top: 0;
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_select.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_select.scss
@@ -4,7 +4,6 @@
   font-size: 0.875rem;
   margin-top: $sizeS;
   width: 100%;
-  min-width: 65px;
 
   @include modifier('inline') {
     margin-top: 0;
--- a/packages/react-search-ui-views/src/ResultsPerPage.js
+++ b/packages/react-search-ui-views/src/ResultsPerPage.js
@@ -8,7 +8,14 @@ const setDefaultStyle = {
   option: () => ({}),
   control: () => ({}),
   dropdownIndicator: () => ({}),
-  indicatorSeparator: () => ({})
+  indicatorSeparator: () => ({}),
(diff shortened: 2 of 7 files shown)


packages/react-search-ui-views/src/styles/themes/reference-ui/components/_select.scss
```diff
@@ -4,7 +4,7 @@
   font-size: 0.875rem;
   margin-top: $sizeS;
   width: 100%;
-  min-width: 60px;
+  min-width: 65px;
```
yakhinvadim: I'm concerned that since we hardcode the width here, we could eventually face an issue where a font with even wider letters would cause ellipsis to appear. Is it possible to make a content define the width of select?

constancecchen: I'm poking around at this now and it looks like it's `react-select`'s default styles/css causing this behavior, which is a bit disappointing. For whatever reason they're absolutely positioning the select text (even when they already have `display: flex` and `align-items: center`going on) and also manually setting text-ellipses, overflow-hidden, etc.

I think my preference would be to override/unset react-select's styles, and then tweak their padding/margins, like so:



Which would then allow the select box to grow normally to accommodate all widths & fonts without truncating.

constancecchen: OK, I think I something working locally for this.

https://github.com/elastic/search-ui/blob/4d555722b6c3fe15f15fb6bc8aee54ac7b893297/packages/react-search-ui-views/src/ResultsPerPage.js#L5-L10

can be expanded to include:

```js
const setDefaultStyle = {
  option: () => ({}),
  control: () => ({}),
  dropdownIndicator: () => ({}),
  indicatorSeparator: () => ({}),
  singleValue: provided => {
    // Pulling out CSS that we don't want
    const { position, top, transform, maxWidth, ...rest } = provided;
    return { ...rest, lineHeight: 1, marginRight: 0 };
  },
  valueContainer: provided => ({ ...provided, paddingRight: 0 })
};
```

Which allows us to get rid of the `min-width` CSS above and generates a more responsive select, like so.

JasonStoltz: Oh wow, thank you @constancecchen. Providing a fix saved me a huge headache.

I put the change in place, mind taking a look?

constancecchen: I can confirm this looks great to me both locally and on storybook! Will approve here once tests are passing!

Issue #93, opened 2019-02-15
This library should work with a "strict" Content Security Policy


JasonStoltz:
For reference: https://github.com/elastic/eui/issues/1431

JasonStoltz:
I did the following to test this. Testing was performed in the "examples/sandbox" project.

I added the following policy in a meta tag in `index.html`.

```
<meta
      http-equiv="Content-Security-Policy"
      content="base-uri 'self'; object-src 'none'; script-src 'self' ; style-src 'self'; style-src-elem 'self' https://fonts.googleapis.com 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='"
    />
```

And ran it with:
```
cd examples/sandbox
npm run build
npm install -g serve
serve -s build
```

As you can see in the policy above, I had to make the following adjustments:

The first thing I had to do was add `https://fonts.googleapis.com` as a trusted domain for `style-src-elem`, since our stylesheet imports styles from there. I think this is OK.

The second thing I had to do was add a hash for `style-src-elem`, since it appears that our dependency, `react-select`, adds inline styles to the `dom` using the `emotion` library.

I believe the second issue will need to be addressed. Either by:

A) Removing the dependency
B) Disable styles being injected by the dependency and provide all of our own styles

yakhinvadim:
I'd like to reopen this.
Looks like the fonts have been fixed already in #293
And react-select issue can be solved by updating it to v3:  JedWatson/react-select#3585 (search for "CSP").

JasonStoltz:
Updating to react-select v3 will likely be a breaking change, as the markup and classnames will likely be differently.

If we do this, I think we need to take the approach of deprecating the old component and adding a new component which can be opted into.

This *could* be a global configuration for Search UI, one that conditionally loads v3 instead of v2.

PR #311, opened 2019-07-02
Make the “+ More” action for showing more filters/checkboxes keyboard accessible

### LINKED ISSUE #307: The “+more” action for showing more filters/checkboxes is not keyboard accessible
This is on the "Facet" component.

constancecchen:
## Description

On the Facet/MultiCheckboxFacet components, the `+ More` action for showing more filters/checkboxes was previous not accessible to keyboard users. This PR addresses that accessibility issue by converting the action to a button (see below screenshot).

## List of changes

- Converts MultiCheckboxFacet's `+ More` action from a `<div>` to a `<button>`
- Adds hover/focus styling to the `+ More` action

## Associated Github Issues

Fixes #307


constancecchen:
⚠️ This is an extra/nice-to-have, but I thought I'd quickly spike making the "More" action more screen-reader-friendly as well:

voiceover

My goal was to create screen reader behavior similar to [inclusive-component's](https://inclusive-components.design/a-todo-list/) [To Do List example](https://codepen.io/heydon/pen/VpVNKW), where after any action (e.g. adding a new item, deleting an item), the action that you just performed gets read out to the user. That way visually impaired users aren't left wondering what happened after pressing a button.

[You can check out the diff/extra code **here**](https://github.com/constancecchen/search-ui/compare/keyboard-accessible-more...constancecchen:screenreader-accessible-more).

I kept it separate from this PR for now (since it mostly addresses keyboard accessibility), but I can merge it in if you feel comfortable @JasonStoltz.

JasonStoltz:
@constancecchen I like your separate changes. The *only* thing I'm thinking about, is custom styling. We are hiding that div in our own stylesheet, but users will also have to know to do that in their custom stylesheets.

I wonder if it would make sense to have a specific class name that indicates that something should be hidden for accessibility, then mention it in a styling guide?

constancecchen:
@JasonStoltz Oh man, excellent catch/point! I'm actually slightly leaning towards making the visually-hidden styling inline in that case, since there is basically no use case where developers/users would _want_ that section to be visible. Would you be cool with that?

JasonStoltz:
@constancecchen Well, here's the only catch. I'm trying to make this library work with a strict CSP: https://github.com/elastic/search-ui/issues/93. These policies will disallow inline CSS by default. 

However, I *think* there are reasonable ways that users can possibly work around this and we already have some inline css from react-select, so it's a larger problem we'll need to tackle at some point anyway.

So yes, just thinking out loud there, I think I'm leaning towards the inline css approach. It's probably fine for now and we could always change it later when we address CSP as a whole.

constancecchen:
Oh wow, that's super fascinating. I wonder how CSS-in-JS libraries like emotion get around this (or maybe they don't address CSP?)

I'm seeing [using CSSOM as a workaround](https://stackoverflow.com/a/27088213/4294462) for getting around the inline-style warning - would that be something you're good with doing as a workaround?

The more I think about this also, the more I think I want to tackle the screen reader announcements in a single live area that's present just once in the app, rather than repeating it multiple times in various locations. For that reason I might merge this PR as-is and open a brand new one for screen reader accessibility/announcements that uses the CSSOM method of setting the visually hidden styles.

JasonStoltz:
@constancecchen Sounds good to me. Also, I try to squash + merge PRs on Github for this project to try to keep a clean git history. Feel free to merge at your leisure.

PR #318, opened 2019-07-09
Fix issue where paging info would show a higher end than possible

constancecchen:
## Description

Caught this issue while doing some accessibility testing! It's pretty minor but I think it'll reduce confusion slightly.

### Before

(Notice 41 - **80**, when the results max/cap out at 59) 

### After

(Using the correct results max)

## List of changes
- Adds a `Math.min()` check to obtain the lesser of `end` and `totalResults`


constancecchen:
I also added a storybook example (I forgot storybook was a thing, sorry! 🤦‍♀) although I'm happy to remove if we don't think it's necessary!

PR #319, opened 2019-07-09
[a11y] More accessible MultiCheckboxFacets

constancecchen:
## Description

This is a quick semantic markup change which enhances many screen readers' ability to associate multiple checkbox groups with its parent group label.

### After

Notice how California now reads out the parent "States" label, giving more context to keyboard tabbing users.

### Screencap

fieldset

^ Posting for general edification - note how the parent `legend` is only read out when the user tabs into the **first** input within the `fieldset` group, and is not read out again when tabbing to other items within the group. This is great behavior for context while not becoming repetitive.

## List of changes

- Changes MultiCheckboxFacet markup from `div`s to `fieldset`/`legend` tags.

## QA

- [x] Tested in MacOS Voiceover
- [x] Tested in Chrome for no visual regressions
- [x] Tested in Firefox for no visual regressions
- [x] Tested in Edge for no visual regressions

### DIFF
--- a/packages/react-search-ui-views/src/MultiCheckboxFacet.js
+++ b/packages/react-search-ui-views/src/MultiCheckboxFacet.js
@@ -19,13 +19,13 @@ function MultiCheckboxFacet({
   searchPlaceholder
 }) {
   return (
-    <div
+    <fieldset
       className={appendClassName(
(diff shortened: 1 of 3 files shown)


constancecchen:
Rebased to fix a merge conflict with #311!

PR #320, opened 2019-07-09
[a11y] Adding a screen reader live region and announcements for actions

constancecchen:
## Description

This is an an accessibility PR that provides more context and information to screen reader users. The goal is to announce the results/consequences of certain actions being taken to users, so that visually impaired users aren't left wondering what happened after pressing a button.

For example:
- After clicking the "+ More" button in the filters sidebar, a screen reader user should know how many total filters are now being shown. (This was originally spiked out [here](https://github.com/elastic/search-ui/pull/311#issuecomment-507883906)).
- After pressing the search button, a screen reader user should know how many results are being shown and how many total results were returned.
- After navigating to another page of results, a screen reader user should have the results/paging info read out to them (same as above)
- After changing how many results are being shown per-page, a screen reader user should have the results/paging info read out to them (same as above)

## Screencaps

`Search results announcements:`

search-results-screencap

`+ More Filters annoucements:`

more-filters-screencap

## List of changes

I recommend following along commit-by-commit if possible!

- Adds a new ScreenReaderStatus render prop helper
  - This creates a one-time live region in the app, and reads out any updates in that region to screen readers. This region can be used by multiple different components.
  - Note: Because this directly modifies the client DOM, a check needs to be added to this helper to ensure it doesn't crash for SSR React apps.
- Updates MultiCheckboxFacet's "+ More" button to announce shown filters to ScreenReaderStatus
- Updates PagingInfo to announce changes in results shown to ScreenReaderStatus

## QA

- [x] Tested on MacOS Voiceover

(Note - uncovered lines are same as before, and all coverage %s went up slightly)

Screen Shot 2019-07-09 at 12 38 10 PM

## Future

I'll continue to investigate other opportunities to use this new live region (the next likely being the SingleLinksFacet component), but for now I'd like to keep this PR fairly small and to the two lowest hanging fruit for screen reader announcements.

### DIFF
--- a/packages/react-search-ui-views/src/MultiCheckboxFacet.js
+++ b/packages/react-search-ui-views/src/MultiCheckboxFacet.js
@@ -3,7 +3,11 @@ import React from "react";
 import deepEqual from "deep-equal";
 
 import { FacetValue, FilterValue } from "./types";
-import { appendClassName, getFilterValueDisplay } from "./view-helpers";
+import {
+  appendClassName,
+  getFilterValueDisplay,
+  ScreenReaderStatus
+} from "./view-helpers";
 
 function MultiCheckboxFacet({
   className,
@@ -12,6 +16,7 @@ function MultiCheckboxFacet({
   onRemove,
   onSelect,
   options,
+  optionsCount,
   showMore,
   values,
   showSearch,
@@ -82,14 +87,28 @@ function MultiCheckboxFacet({
       </div>
 
       {showMore && (
-        <button
-          type="button"
-          className="sui-multi-checkbox-facet__view-more"
-          onClick={onMoreClick}
-          aria-label="Show more options"
-        >
-          + More
-        </button>
+        <ScreenReaderStatus
+          render={announceToScreenReader => (
+            <button
+              type="button"
+              className="sui-multi-checkbox-facet__view-more"
+              aria-label="Show more options"
+              onClick={() => {
+                onMoreClick();
+
+                const newLimit = options.length + 10;
+                const showingAll = newLimit >= optionsCount;
+                const message = `${
+                  showingAll ? `All ${optionsCount}` : newLimit
+                } options shown.`;
+
+                announceToScreenReader(message);
+              }}
+            >
+              + More
+            </button>
+          )}
+        />
       )}
     </fieldset>
   );
@@ -102,6 +121,7 @@ MultiCheckboxFacet.propTypes = {
   onSelect: PropTypes.func.isRequired,
   onSearch: PropTypes.func.isRequired,
   options: PropTypes.arrayOf(FacetValue).isRequired,
+  optionsCount: PropTypes.number,
   showMore: PropTypes.bool.isRequired,
   values: PropTypes.arrayOf(FilterValue).isRequired,
   className: PropTypes.string,
--- a/packages/react-search-ui-views/src/PagingInfo.js
+++ b/packages/react-search-ui-views/src/PagingInfo.js
@@ -1,17 +1,27 @@
 import PropTypes from "prop-types";
 import React from "react";
 
-import { appendClassName } from "./view-helpers";
+import { appendClassName, ScreenReaderStatus } from "./view-helpers";
 
 function PagingInfo({ className, end, searchTerm, start, totalResults }) {
+  end = Math.min(end, totalResults);
   return (
-    <div className={appendClassName("sui-paging-info", className)}>
-      Showing{" "}
-      <strong>
-        {start} - {Math.min(end, totalResults)}
-      </strong>{" "}
-      out of <strong>{totalResults}</strong> for: <em>{searchTerm}</em>
-    </div>
+    <>
+      <div className={appendClassName("sui-paging-info", className)}>
+        Showing{" "}
+        <strong>
+          {start} - {end}
+        </strong>{" "}
+        out of <strong>{totalResults}</strong> for: <em>{searchTerm}</em>
+      </div>
+      <ScreenReaderStatus
+        render={announceToScreenReader => {
+          let message = `Showing ${start} to ${end} results out of ${totalResults}`;
+          if (searchTerm) message += `, searching for "${searchTerm}".`;
+          return announceToScreenReader(message);
+        }}
+      />
+    </>
   );
 }
 
--- a/packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
+++ b/packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
@@ -41,6 +41,7 @@ const getLiveRegion = () => {
 const announceToScreenReader = announcement => {
   const region = getLiveRegion();
   region.innerText = announcement;
+  return null;
 };
 
 /**
--- a/packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
+++ b/packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
@@ -6,6 +6,7 @@
  * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
  */
 const regionId = "search-ui-screen-reader-announcements";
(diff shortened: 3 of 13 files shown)


packages/react-search-ui-views/src/PagingInfo.js
```diff
@@ -1,17 +1,27 @@
 import PropTypes from "prop-types";
 import React from "react";
 
-import { appendClassName } from "./view-helpers";
+import { appendClassName, ScreenReaderStatus } from "./view-helpers";
```
constancecchen: One potential downside to adding this to the `<PagingInfo />` component is that if users roll their own view, they're inadvertently making their app less accessible - so it's possible that adding this to the `react-search-ui` container may be a bit more robust. Definitely open to thoughts on this!

packages/react-search-ui-views/src/MultiCheckboxFacet.js
```diff
@@ -82,14 +87,28 @@ function MultiCheckboxFacet({
...
+        <ScreenReaderStatus
+          render={announceToScreenReader => (
+            <button
+              type="button"
+              className="sui-multi-checkbox-facet__view-more"
+              aria-label="Show more options"
+              onClick={() => {
+                onMoreClick();
+
+                const newLimit = options.length + 10;
```
constancecchen: This is a bit of a magic constant (from the parent [Facet.js container](https://github.com/elastic/search-ui/blob/9086bcf5ed438dbbf4d486080b22c704442cf587/packages/react-search-ui/src/containers/Facet.js#L50)) - feel free to shout if you think it should be converted to a prop or something a bit less static

packages/react-search-ui-views/src/PagingInfo.js
JasonStoltz: In this project it's often a bit fuzzy to me what should go into a container and what should go into a view. 

The guidance we give users for customizing views is to use the default view as a base and customize from there. That could be a bit of a foot gun if they forget to call "announceToScreenReader", as you point out.

The goal of having swappable views, is to give users a very easy way to use custom markup for a component. The less they have to reproduce in the view, the better. To that end, it would seem more ideal to do this in a container. We could either do it completely transparently, or we could roll all of the logic into a single prop that we pass down.

```jsx
function PagingInfo({ className, end, searchTerm, start, totalResults, screenReaderAnnouncement }) {
  end = Math.min(end, totalResults);
  return (
    <>
      <div className={appendClassName("sui-paging-info", className)}>
        Showing{" "}
        <strong>
          {start} - {end}
        </strong>{" "}
        out of <strong>{totalResults}</strong> for: <em>{searchTerm}</em>
      </div>
      {screenReaderAnnouncement(someCustomMessageFunction)}
    </>
  );
}
```

It could be something like, if you don't call it at all, the default announcement will be used. if you call it with a custom function, `screenReaderAnnouncement()`, then that overrides the default message with a custom message.

JasonStoltz: I also envisioned a future state where users could use `react-search-ui` without including `react-search-ui-views` at all, so they could have a completely custom view layer. Having this be part of the container would make that a challenge 🤔.

packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
```diff
@@ -0,0 +1,61 @@
...
+  region.style.border = "0";
+  region.style.overflow = "hidden";
+  region.style.clip = "rect(0 0 0 0)";
+
+  document.body.appendChild(region);
+  return region;
+};
+
+const announceToScreenReader = announcement => {
+  if (hasDOM) {
```
JasonStoltz: Do we need to do explicit DOM manipulation? Could this a more declarative, React type thing? Like could announcements be managed in central state somewhere? 

For instance, messages could go into their own context store and we could just wire a ScreenReaderStatus component up to that context and re-render when it changes.

We could even take it a step further and make screen reader status part of the core Search UI state, and have an action to update it.

JasonStoltz: If it were part of the core Search UI state, it would make it available for any library also, whether it's Vue, React, or vanilla js, etc.

JasonStoltz:
Also 💯 for this PR.

packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
constancecchen: At some point we'll probably want to do explicit DOM manipulation (we could use a React portal, but that's also essentially just DOM appending/manipulation and will still need a document/window check for SSR apps). Also FWIW, I looked at [how downshift was doing announcements](https://github.com/downshift-js/downshift/blob/master/src/set-a11y-status.js) as well after noticing their live region, and they're essentially doing what we're doing, which gives me a bit of confidence.

constancecchen: Ah, wait, no, I think I maybe see what you mean by central state. Originally when I was making this, I was really only thinking in terms of putting this in `react-search-ui-views`, and my hesitation towards putting it in, e.g. `<Layout>` was something along the lines of "What if the developer isn't using our `Layout` component".

But if we're moving to `react-search-ui` instead (correct me if I'm wrong), then we can take advantage of the top level `<SearchProvider>` component itself to create the region + manage a central state.

Is that what you were thinking?

constancecchen: Or, wait, sorry, I'm still catching up again :) You're thinking even higher level than that - putting it in [SearchDriver.js](https://github.com/elastic/search-ui/blob/master/packages/search-ui/src/SearchDriver.js) (or it's own file?) - correct me if I'm wrong?

I'd love to get a chance to chat with you more about this to clarify architecture a bit more!

packages/react-search-ui-views/src/MultiCheckboxFacet.js
```diff
@@ -82,14 +87,28 @@ function MultiCheckboxFacet({
...
-        <button
-          type="button"
-          className="sui-multi-checkbox-facet__view-more"
-          onClick={onMoreClick}
-          aria-label="Show more options"
-        >
-          + More
-        </button>
+        <ScreenReaderStatus
+          render={announceToScreenReader => (
```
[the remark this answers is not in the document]
constancecchen: 1. I've used both in the past, just tend to user `render` out of habit. Good with switching to `children` or whatever we prefer!

2. After reading your comments down below I think I'm actually leaning towards moving away from a render prop (if we do indeed want to make this usable to non-React apps). Re: custom hooks specifically - I ran into crashes/errors when I tried to use hooks with the `MultiCheckboxFacet` component (I _think_ related to the previous hook issues we've seen?) - let me try to create a reproducible branch.

3. Yeah, I definitely didn't think too much about making this work for developers using custom views/not using react views, etc. I think we're starting to think about moving away from this in other comments.

4. See previous comments - I think we're discussing moving away from this being React dependent, correct me if I'm wrong? (although I'm definitely not 100% on this, so happy to get clarification)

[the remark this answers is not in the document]
constancecchen: Re 2. - here's an [example branch](https://github.com/constancecchen/search-ui/tree/hooks-error)/[commit](https://github.com/constancecchen/search-ui/commit/8201c0a5b10e494fcf7acdec064920f75cd9a953) with a very basic hook usage that creates the following console error:

Screen Shot 2019-07-11 at 8 24 18 AM

packages/react-search-ui-views/src/view-helpers/screenReaderAnnouncements.js
JasonStoltz: @constancecchen Yes, I think this warrants a chat. Will message you later today!

JasonStoltz: Summary of our chat: https://github.com/elastic/search-ui/issues/325

constancecchen:
Closing this in favor of a new PR: #355

PR #324, opened 2019-07-11
Facet CSS cleanup

constancecchen:
## Description

While working on facet accessibility, I happened to notice that some of our facet-related CSS was a bit repetitive and could stand to be DRYed out (particularly around styling the facet group labels).

The net benefit is hopefully that overriding styles on all facet groups/facet titles should become much more consistent and predictable.

This does change our className<->DOM structure relationship a bit (I opted on the side of wanting to simplify things more before we chatted about whether that constituted as a breaking change today) - but if strongly preferred, I can move the facet group CSS blocks back to where they were - just let me know.

## List of changes

I recommend following along commit-by-commit if possible!

- Refactored SingleSelectFacet to reuse existing `.sui-facet` & `.sui-facet__title` styling, and removed unused CSS
- Refactored MultiCheckboxFacet to reuse existing `.sui-facet` & `.sui-facet__title` styling, and removed unnecessary flex CSS
- Refactored SingleLinksFacet to reuse existing `.sui-facet` & `.sui-facet__title` styling, and moved CSS for this facet to its own file (it was not being used by another component as far as I could tell), + removed unused an `--active` modifier

## QA

There should be no visual difference or regression between current master and this branch. This is a code refactor only.

Before & After:


JasonStoltz:
@constancecchen Since this is breaking, we'll to note what the breaking changes are in the Release Notes. Would you mind writing up a quick summary of what exactly someone would need to do who may have been targeting the old classnames when upgrading to this version?

Issue #214, opened 2019-04-19
Add mobile responsiveness


aznick:
Make the default components more mobile friendly.

We could add components tuned for mobile or add responsive transformations to existing components (with a preference for the former).

With the current styles, the experience is quite poor on mobile:

https://search-ui-stable.netlify.com/

<img width="701" alt="React_App" src="https://user-images.githubusercontent.com/1427475/60988282-4641ce00-a311-11e9-8611-b6b13c24dc44.png">

johnbarrierwilson:
Just a wee bit o' documentation from our discussion/meeting...

We discussed many different possible options for making the filters responsive. The most popular pattern that was discussed was the Wayfair-like flyout with accordions for each individual feature.

In order to get from point-A (conceptual discussions) to point-B (fully implemented responsive filters), I'll be working on:
1) Researching the codebase to see what is do-able
2) Create a few conceptual mockups to help decide on best implementation path

JasonStoltz:
Mentioning the associated PR here, since I don't think it was tagged in the PR itself: https://github.com/elastic/search-ui/pull/344

constancecchen:
No strong feelings either way, but we might want to consider leaving this open since #344 was only meant to be an MVP solution to responsiveness, and doesn't fully implement some nicer UX that users might want to see. I'll defer to @aznick and @johnbarrierwilson on that!

JasonStoltz:
I'll defer to ya'll, but I think we should close this, and if we have more specific mobile UX requirements in the future, we'll open a new issue. This was issue was submitted primarily because the mobile experience was terrible. Since the mobile experience is no longer terrible, I consider this closed.

aznick:
Agreed with @JasonStoltz - issues are meant to track a discrete piece of work or thought. (And closing issues feels good 😎). 

But if you have specific ideas about how we could improve the mobile experience, please feel free to open a new issue @constancecchen! It's a good idea to capture those thoughts and let them marinate, even if it isn't an issue for this sprint.

PR #334, opened 2019-07-13
Add basic/MVP mobile UX for layout sidebar

constancecchen:
## Description

From our Slack discussion:

> I thought it'd be a shame to ship 1.0 without very basic mobile functionality - so I spiked out a quick attempt at a mobile toggle for our sidebar/filters. I'm hoping that it's a good enough start for the upcoming release and that we can continue to iterate on it for 2.0+

## Screenshots

screencap

## List of changes

1. Adds a new internal [LayoutSidebar](https://github.com/elastic/search-ui/pull/334/files#diff-0433fdec156966cc8e783042c78eff6c) component + styling which:
  - handles conditionally rendering show/hide toggle buttons
  - makes the filter sidebar a scrollable sticky flyout on mobile
  - does not display toggle buttons if there is no content in the sidebar

Note that for developers that utilize `<Layout>` but are completely rolling their own styles - they will not inherit this mobile UX as a result, but the toggle buttons also should not show up or otherwise mess up their styling.

At some point the goal is to separate layout vs presentational CSS so that users can have both completely custom styling while still having OOTB working mobile UX.

2. Additionally, I also proposed a minor cleanup/refactor in 7cfcd4a which unflattens all our nested classes in `_layouts.scss` - which should make overriding existing styles much more pleasant for developers (`.sui-layout .sui-layout-body .sui-layout-sidebar` becomes just `.sui-layout-sidebar`, for example).

## QA

- [x] Test in Chrome, Safari, Opera
- [x] Test in Firefox
- [x] Test in Edge
- [x] Test in iOS Safari
- [x] Test in Android Chrome

## Associated Github Issues

#214 - although note that this PR does not close/resolve the issue - it's just a v1 and I wouldn't consider this a finished/polished feature by any means.

### DIFF
--- a/packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
+++ b/packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
@@ -1,0 +1,27 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders correctly 1`] = `
+<Fragment>
+  <button
+    className="sui-layout-sidebar-toggle"
+    hidden={true}
+    onClick={[Function]}
+    type="button"
+  >
+    Show Filters
+  </button>
+  <div
+    className="sui-layout-sidebar"
+  >
+    <button
+      className="sui-layout-sidebar-toggle"
+      hidden={true}
+      onClick={[Function]}
+      type="button"
+    >
+      Done Filtering
+    </button>
+    Hello world!
+  </div>
+</Fragment>
+`;
--- a/packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
+++ b/packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
@@ -19,7 +19,7 @@ exports[`renders correctly 1`] = `
       onClick={[Function]}
       type="button"
     >
-      Done Filtering
+      Save Filters
     </button>
     Hello world!
   </div>
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
@@ -45,6 +45,9 @@
       margin: 32px 0;
       padding-right: 32px;
 
+      @media (max-width: $sidebarBreakpoint + $offset) {
+        padding-right: 0;
+      }
       @media (max-width: $sidebarBreakpoint) {
         z-index: 99;
         display: none;
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
@@ -97,6 +97,7 @@
     width: 100%;
     margin: 20px auto;
     padding: 10px;
+    font-family: inherit;
     font-size: 14px;
     font-weight: 700;
     color: $linkColor;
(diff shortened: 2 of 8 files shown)


packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
```diff
@@ -0,0 +1,27 @@
...
+  <div
+    className="sui-layout-sidebar"
+  >
+    <button
+      className="sui-layout-sidebar-toggle"
+      hidden={true}
+      onClick={[Function]}
+      type="button"
+    >
+      Done Filtering
```
johnbarrierwilson: This is nit-picky, but "Save Filters" might be wording as it promotes an actual action that the user takes to modify the action of the computer/phone ("Save") versus a state-of-being that the user achieves.

johnbarrierwilson: Also, I just posted this on the `.test` file.... and I also realize it's Monday.
giphy

packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
```diff
@@ -2,67 +2,143 @@
...
+ */
 
-  @include block("layout-header") {
-    padding: 32px 24px;
-    border-bottom: 1px solid #eeeeee;
+@include block("layout-body") {
+  background: #fcfcfc;
+
+  &:after {
+    content: "";
```
johnbarrierwilson: What are your thoughts on alphabetizing properties? Is there another strategy you use here to keep properties organized? I usually highlight the same-level properties after writing them, and just get the text editor the alphabetize them.

johnbarrierwilson:
Really great work @constancecchen! Thanks for knocking this out.

packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
constancecchen: For this line specifically I don't believe I touched in the CSS itself, just un-nested/un-indented it - but in general, I really prefer @mdo's [Code Guide](https://codeguide.co/#css-declaration-order) which orders by positioning, box model, then typography/visual, etc. This helps highlight properties that can massively impact how CSS displays (e.g. z-index, display, position) while keeping properties grouped that make sense together (e.g. flex properties). The full property order list is [here](https://github.com/twbs/stylelint-config-twbs-bootstrap/blob/master/css/index.js#L54) (which I believe is based on/originated from Twitter's Recess library).

packages/react-search-ui-views/src/__tests__/layouts/__snapshots__/LayoutSidebar.test.js.snap
constancecchen: I'm good with that change! Just to check, does "Save Filters" also make sense for when you're clearing/have zero filters checked? Or should we eventually make a separate "Clear Filters" button for that anyway?

johnbarrierwilson: For now, I think it's fine. I do agree that a "clear filters" button would be great in the future. And when the user clicks "clear filters," it will automatically clear _and_ collapse the filters.

packages/react-search-ui-views/src/styles/themes/reference-ui/layouts/_layout.scss
johnbarrierwilson: Interesting! I've always had trouble with how bootstrap organized properties, but that might be because I've never seen this. I'll look into it some more!

Issue #290, opened 2019-06-26
Update to corejs 3 and remove babel-polyfill


JasonStoltz:
https://babeljs.io/blog/2019/03/19/7.4.0

JasonStoltz:
This is no longer relevant, since the babel-polyfill dependency was removed.

Issue #278, opened 2019-06-26
Error when adding search-ui to next.js project on code sandbox


JasonStoltz:
regeneratorRuntime is not defined
ReferenceError: regeneratorRuntime is not defined
    at /sandbox/node_modules/@elastic/search-ui-app-search-connector/lib/AppSearchAPIConnector.js:135:7
    at /sandbox/node_modules/@elastic/search-ui-app-search-connector/lib/AppSearchAPIConnector.js:176:6
    at Object.<anonymous> (/sandbox/node_modules/@elastic/search-ui-app-search-connector/lib/AppSearchAPIConnector.js:241:2)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:690:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/sandbox/node_modules/@elastic/search-ui-app-search-connector/lib/index.js:263:53)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)

JasonStoltz:
To work around this, I had to `import "@babel/polyfill";` at the very top of my fie before importing search-ui.

JasonStoltz:
One thought, is that I am not including this in the entrypoint for the `react-search-ui` library, which is causing the issue.

JasonStoltz:
Actually, I think it's the other way around:

It happens on this import:

import AppSearchAPIConnector from "@elastic/search-ui-app-search-connector";

But not on any of these:

import {
 SearchProvider,
 Results,
 SearchBox,
 ErrorBoundary,
 WithSearch
} from "@elastic/react-search-ui";
import { Layout } from "@elastic/react-search-ui-views";

Maybe I'm just doing something wrong in the app search conector

FokkeZB:
[...]

PR #340, opened 2019-07-15
Removed babel polyfill

JasonStoltz:
## Description

`@babel/polyfill` was a dependency in this project. This was a pretty crazy headache, as it was a runtime dependency (as opposed to a build time, dev dependencies) that actually goes out and Polyfills a number of global APIs, like `Promise`, `Map`, etc, which is very intrusive and there is a lot of potential for conflict there. It's especially bad since this is a library, and you would *not* expect a library to do that.

In Babel's docs, they actually outline two methods for handling these globals. The first is the Polyfill, which they advise against for the libraries. The second, is https://babeljs.io/docs/en/babel-plugin-transform-runtime, which they DO recommend for libraries.

The approach the plugin takes is to replace globals like "Promise.all" etc with helpers from the `babel-runtime` package. This is much safer for libraries as it's not actually shimming anything globally.

Initially, I attempted to remove `@babel/polyfill` and not replace it, but it is required since we use `async` functions, which require a regenerate runtime.

Long story short, I replaced `@babel/polyfill` with `@babel/plugin-transform-runtime`

## List of changes
- Removed `@babel/polyfill`
- Added `@babel/plugin-transform-runtime`

## Associated Github Issues
https://github.com/elastic/search-ui/issues/278
https://github.com/elastic/search-ui/issues/290

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,7 +122,7 @@ It can often be useful to publish a change and test it before doing a real live 
 option for this.
 
 1. Check out pr branch
-2. `npx lerna publish --canary` - Publishing with canary generates a unique new version number, publishes it to npm, then updates the `canary` tag in npm to point to that new version.
+2. `npx lerna publish --canary --preid canary` - Publishing with canary generates a unique new version number, publishes it to npm, then updates the `canary` tag in npm to point to that new version.
 3. In project you'll need to install all dependencies explicitly with the canary tag from npm:
 
    ```
--- a/examples/elasticsearch/package-lock.json
+++ b/examples/elasticsearch/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "elasticsearch-example",
-  "version": "0.12.1",
+  "version": "1.0.0-rc.0",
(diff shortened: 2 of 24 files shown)


CONTRIBUTING.md
```diff
@@ -122,7 +122,7 @@ It can often be useful to publish a change and test it before doing a real live
 option for this.
 
 1. Check out pr branch
-2. `npx lerna publish --canary` - Publishing with canary generates a unique new version number, publishes it to npm, then updates the `canary` tag in npm to point to that new version.
+2. `npx lerna publish --canary --preid canary` - Publishing with canary generates a unique new version number, publishes it to npm, then updates the `canary` tag in npm to point to that new version.
```
constancecchen: This command is starting to become a bit of a mouthful - do we want to create a package.json `scripts` shorthand for this? something like `canary`, so devs can just type `npm run canary` or `yarn canary` instead?

JasonStoltz: Yes, that is a great idea. I can do that in a follow up PR.

PR #341, opened 2019-07-15
Fix misc. Edge cross-browser styling issues

constancecchen:
## Description

This fixes some very minor cross-browser Edge CSS issues that I noticed while testing Search UI's sandbox example in Browserstack.

### Screenshots

**Before (left) vs. After (right):**

(Weird squashing/padding on right appeared to be due to flexbox & shrinking)

(Very subtle sans-serif vs. Segoe font in "Filter states")

## List of changes

- Fixes font family on SearchBox's text/button input not inheriting correctly on Edge
- Fix SearchBox's "Search" button being squashed on Edge
- Fixes font family on Search Filters input not inheriting correctly on Edge

### DIFF
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_multi-checkbox-facet.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_multi-checkbox-facet.scss
@@ -55,6 +55,7 @@
     height: 100%;
     padding: 6px;
     margin: 0;
+    font-family: inherit;
     border: 1px solid #cccccc;
     border-radius: 4px;
     outline: none;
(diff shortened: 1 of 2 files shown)


yakhinvadim:
The PR looks good, but I'm curious, why does Edge need font-family: inherit? Isn't it the default value?

constancecchen:
Sadly form elements are annoyingly opinionated and tend to come with default browser styling/overrides. I don't exactly know why this was working on Chromium/FF but not Edge to be honest with you - TBH I would have expected Chrome/FF/etc. to also have the incorrect font 🤷‍♀ Sometime in the distant future this might be a moot point anyway since Edge is moving to the Chromium engine.

Couple related links/resources - [Normalize.css for example](https://github.com/necolas/normalize.css/blob/master/normalize.css#L165) resets font-family for both inputs and buttons, and [css-tricks has a pretty good writeup on default button styles here](https://css-tricks.com/overriding-default-button-styles/).

PR #343, opened 2019-07-15
Updated CONTRIBUTING guide

JasonStoltz:
- Updated publish commands to include "force-publish", so that ALL
packages are published, regardless of whether changes are made.
- Changed `from-git` to `from-package` in case someone makes a CHANGELOG
change before publishing.
- Added specific commands for releasing RCs
- Added links to code sandbox examples


CONTRIBUTING.md
```diff
@@ -131,6 +131,26 @@ option for this.
 
 4. To Deploy, simply push your changes to the `canary` branch, then visit "https://search-ui-canary.netlify.com/"
 
+### Release candidates
+
+When pushing release candidates, the following lerna commands can be useful.
```
goodroot:
```suggestion
When pushing release candidates, the following lerna commands can be useful:
```

CONTRIBUTING.md
```diff
@@ -131,6 +131,26 @@ option for this.
...
+```
+
+### Testing Canary build and Release Pre-releases
+
+We have a number of demos available that you can use to do quick smoke testing of releases in various
+stacks:
+
+https://codesandbox.io/s/search-ui-gatsby-example-u041m
+https://codesandbox.io/s/search-ui-next-js-example-tb05u
+https://codesandbox.io/s/search-ui-national-parks-example-kdyms
```
constancecchen: I think this needs to have `- ` bullet point prefixes - right now it's just outputting a single line of links in a row which is a bit hard to click.



The stable demos section/links below would probably benefit from the same.

JasonStoltz: Oh yes, you're right. Thank you :)

JasonStoltz:
@constancecchen Addressed your feedback and also added a couple of other changes.

PR #342, opened 2019-07-15
Revert "Add basic/MVP mobile UX for layout sidebar (#334)"

JasonStoltz:
This reverts commit b0b3cf46669dfd484c756ad1ae2f294e11a16bd8.

PR #344, opened 2019-07-16
Add basic/MVP mobile UX for layout sidebar (v2)

### LINKED ISSUE #214: Add mobile responsiveness

constancecchen:
## Description

See #334 for original PR and and #342 for reversal.

This is essentially #334 but with https://github.com/elastic/search-ui/pull/334/commits/c2baba269d1751c55fea073003731a201f5de103 ("_does not display toggle buttons if there is no content in the sidebar_") struck off. So the toggle buttons are always there on mobile now, regardless of sidebar content or whether search results are present.

I've tried for a few hours now but I can't seem to find a sane and SSR-friendly way of detecting whether the sidebar has content (I've tried refs, querying the search params, etc.), and so at least for the MVP, I'd suggest defaulting to my initially suggested Option 1 - which is shipping as-is. It's not the end of the world to have an empty flyout, and it's still a better worst case scenario than not having a working UI/UX for mobile users at all.

## QA

I've tested this in the following client apps and can confirm it's working and no longer crashing with errors:

- [x] Gatsby
- [x] Create React App
- [x] next.js

Closes #214


JasonStoltz:
@constancecchen One thing to think about. *If* we consider this a breaking change, would there be a way to make it "opt-in" behavior, so we do not break existing implementations? That way, we could release it in a 1.1, rather than a 2.x.

constancecchen:
Visually, this should not be a breaking change at all. This is brand new code, so there should be no existing user styles/overrides for the button that could break, and for users who are rolling their own custom styles and not importing Search UI, the buttons degrade gracefully to being hidden and behavior should be the same as before (whatever users were doing by themselves).

Is there another definition of breaking change that you're looking for that I'm missing?

yakhinvadim:
If some users implemented that functionality themselves, will they end up with two buttons and two sidebars after updating?

constancecchen:
They will not if they don't add their own styling for `.sui-layout-sidebar-toggle` that sets `display: block` in CSS (and there would be no reason for them to do so if they're rolling their own styles and markup). Our buttons will remain hidden unless deliberately re-displayed via CSS.

yakhinvadim:
Yes, that makes total sense, Constance! 
Now I understand that it wouldn't change anything visually for those who use their own markup or their own styles.

But I'm still concerned about the users who use both our markup and styles. If they have a screenshot testing set up, this change would break their tests, right? If this is the case, 
I don't think we should bend the semver rules here and push it as is in 1.1. I see two options here:
1) make it opt-in as Jason suggested,
2) release a new major version in the nearest future. This would not be that terrible I guess since we'll have to release a 7.3/7.4 to have the same version as Elastic Stack anyway. 

Please let me know if I missed anything.

constancecchen:
> But I'm still concerned about the users who use both our markup and styles. If they have a screenshot testing set up, this change would break their tests, right?

For **screenshot** testing (assuming you mean visual diffing) - no, this shouldn't break tests for users using their own styles. (If they're using our styles, that would be somewhat baffling as to why they would be testing a 3rd party dependency's styling).

For **snapshot** testing - I don't think this would break either. Snapshots don't dive into child components, so a developer snapshotting our `<Layout>` component would still only see `<Layout>` (plus whatever props they're passing into it). Again, users shouldn't, for the most part, be attempting to test 3rd party dependencies, so they shouldn't ever be diving deeply enough into `Layout` to ever see `<LayoutSidebar>`. Also to be honest snapshots feel very easy to update (just a simple `jest -u`), so I wouldn't lean towards that being considered a "breaking" change.

That being said I'd be fine punting this off until the next major release, no strong feelings there!

JasonStoltz:
The thing I was thinking that I thought would make this potentially a breaking change, is if someone is applying styles on TOP of our css. Like if they are explicitly trying to override styles that we have in place, and now there are additional styles that they haven't accounted for, it *could* mess with their layout. 

Like if you have CSS changes that target ANY existing classnames, then this is potentially breaking.

If we could make this an opt-in thing, like `<Layout mobile="true">`, which would swap out the old html with new html, or add a new className, then I think we could safely call them non-breaking.

Does that make sense?

constancecchen:
Hm, I'm actually still not seeing where users currently targeting out existing layout styles (i.e. `.sui-layout-sidebar`) will suffer breaking changes.

This PR primarily _creates_ new markup and classes (`.sui-layout-sidebar-toggle` and `.sui-layout-sidebar--toggled`) but it should not affect existing desktop overrides (styles on top of our styles). Users couldn't possibly be already targeting the aforementioned two new classes that don't even exist yet. I suppose in theory someone could be doing `.sui-layout-sidebar > *:first-child`, but the odds of that seem infinitesimally small. To help me understand better, is there an example either of you could give of how this would break a current user's styling?

All that being said - I do feel moderately strongly that I'd rather this be a breaking 2.0 release than a 1.x with a flag. In theory the correct and inifintely cleaner way to opt out should be rolling your own styling and layout :)

JasonStoltz:
Hey @constancecchen.

So like this, for example:

Old:
```
@include block("layout-main") {
  width: 76%;	  
  padding: 2rem 0;
  padding: 32px 0 32px 32px;
  border-left: none
```

New:
```
@include block("layout-main") {
  width: 76%;
  margin: 32px 0;
  padding-left: 32px;
```

If I had done something like this in my CSS, with the intent to remove all of the space around my layout:
```
.sui-layout-main {
  padding: 0;
}
```

When I pull down this version, we are now applying a 32px margin, which I have no accounted for, so my site will visually change. I would have to change my CSS in order to correct that:

```
.sui-layout-main {
  padding: 0;
  margin: 0;
}
```

Does that sound right?

constancecchen:
Ahhh, gotcha gotcha, I see - thanks so much for the helpful example! 🙇‍♀ For that specific case, I was being a bit opinionated CSS-wise - if that's the primary instance, I can certainly revert this back to its original padding (vs. margin).

However, I'm also okay just waiting for the 2.0 release. Totally up to you!

JasonStoltz:
@constancecchen My preference is to try to get this out in a minor release. It's a fantastic feature, and I'd much rather release it rather than having it sit in a branch until a 2.0.

constancecchen:
Fair enough! Between the two options of reverting CSS changes to existing classes and adding a flag, which do you prefer?

I'll add that I think I'd strongly prefer the former to adding a flag, but that's just my personal preference :)

JasonStoltz:
@constancecchen That is fine with me!

constancecchen:
@JasonStoltz Alrighty, the opinionated margin/padding changes have been reverted in c244422.

One quick thing to note - I kept the removal of [this extra padding line](https://github.com/elastic/search-ui/pull/344/files#diff-2b7e7b58afbb333a62bdee1fc670a4f8L41) (which was being overridden by the next line anyway) and [this border-left property](https://github.com/elastic/search-ui/pull/344/files#diff-2b7e7b58afbb333a62bdee1fc670a4f8L43) (which wasn't doing anything at all), which I don't think is breaking, but let me know if you disagree.

I didn't see anything else that wasn't either brand new CSS or wasn't in a `@media` query - but please let me know if I missed something and there's another change you'd consider breaking in there!

JasonStoltz:
@constancecchen I approve of these changes. I canaried these changes so that @aznick can give them a look before we merge: https://search-ui-canary.netlify.com/?q=park&size=n_20_n.

aznick:
LGTM - [slack convo](https://elastic.slack.com/archives/CDW125ZRQ/p1563887803023400)

PR #353, opened 2019-07-17
Automatically bake in subfolder installs into `npm install`

constancecchen:
## Description

I kept messing up dependency installs (documented [here](https://github.com/elastic/search-ui/blob/master/CONTRIBUTING.md#installing-dependencies)) and thought I'd make it more idiot-proof for future me 😄

Now instead of having to run 3 separate install commands, just running the default `npm install`/`yarn install` commands will also trigger installing `packages/package.json` and `lerna bootstrap`.

## List of changes

- Add a postinstall script that auto-installs subfolder dependencies
- Update CircleCI
- Update docs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,11 +34,5 @@ jobs:
             - node_modules
           key: v2-dependencies-{{ checksum "package.json" }}
 
-      # Install common deps
-      - run: (cd packages && npm install)
-
-      # Install deps in all packages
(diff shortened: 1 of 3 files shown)


constancecchen:
Phew, looks like CircleCI does indeed run `postinstall` after `npm ci` (was a little worried about that) - merging! 🎉

yakhinvadim:
I really like postinstall usage here 👍

PR #354, opened 2019-07-17
More robust fix for PagingInfo showing a higher end than possible

constancecchen:
## Description

It's https://github.com/elastic/search-ui/pull/318 redux! 😅 When working in `react-search-ui`'s PagingInfo component, I finally got my mind out of "views" mode and realized that we need to fix the issue at the container level and not the view level - since we're still sending the incorrect "end" to users using custom views. Also, the a11y announcement (that I'm working on) needs the correct numbers as well.

## List of changes

- Removes old `Math.min()` code and moves the `totalResult` check logic to `PagingInfoContainer` instead
- Updates tests


constancecchen:
Awesome! Per our ongoing branching discussion in Slack, it sounds like we're keeping our current branching system (where `master` is essentially gitflow's `develop`) so I'll go ahead and merge this in as-is.

Issue #325, opened 2019-07-11
Add screen reader live region support


JasonStoltz:
This was proposed in the following PR: https://github.com/elastic/search-ui/pull/320.

After discussing, we believe screen reader live region support should be supported at the core level of search-ui, not just as part of the view.

To achieve that, we will do the following:

1. We will write notifications to a live region from SearchDriver, so that regardless of what view you are using, updates will always be written.
2. Notifications will be written as a side-effect of triggering actions. For example, after each new search request, we will write a message to the live region, something to the effect of 

    > "Showing 1 to 17 results out of 17, searching for "lake".

    This will be read aloud by the screen reader as updates occur.
3. Notification writing will be encapsulated in a helper, similar to `URLManager`. The helper will be doing explicit DOM manipulation in a DOM element that it will append to the body. This element is completely invisible and should be completely unobtrusive to users and developers.
4. The default behavior of the `SearchDriver` will be to write these live region updates directly to the DOM. However, they can be disabled by passing a boolean flag to the `SearchDriver` configuration. This also applies to the `SearchProvider` component, which simply passes configuration through to the `SearchDriver`.

    ```js
    SearchDriver({
      a11yNotifications: false
    })
    ```

    ```jsx
    <SearchProvider config={{ a11yNotifications: false }} />
    ```

5. In order to support internationalization concerns and customization, the notifications that are written to the live region should be configurable. They should be functions, rather than static strings, so that a developer can handle plurals ("result" vs "results").

    ```js
    SearchDriver({
      a11yNotificationMessages: {
        newSearchResults: ({searchTerm, begin, end, total}) => `Showing ${begin} to ${end} results out of ${total}, searching for "${searchTerm}"`.
      }
    })
    ```  

    Passing nothing for a message, omitting a message, or omitting the configuration entirely will use the default message for message.
    (Optional requirements) Passing `false` could disable that notification entirely

6. An action should be passed down that allows explicit updating of the live region. Something like `a11yNotification`.

    Example: If someone created a custom "RecommendResults" component, they could wire it up to call `a11yNotification`. Something like `a11y("Recommended results: here are some results...")`

PR #355, opened 2019-07-17
Screen reader notifications

constancecchen:
## Description

#320 redux - moving things out of `views` to `SearchDriver` 🚂 

_TODO_ - I'm opening this now to get some sanity checks/discussion points before moving further in my assumptions.

## List of changes

_TODO_

## Associated Github Issues

#325

### DIFF
--- a/packages/search-ui/src/SearchDriver.js
+++ b/packages/search-ui/src/SearchDriver.js
@@ -138,11 +138,10 @@ export default class SearchDriver {
     this.a11yNotifications = a11yNotifications;
     if (this.a11yNotifications) a11y.getLiveRegion();
 
-    a11yNotificationMessages = {
+    this.a11yNotificationMessages = {
       ...a11y.defaultMessages,
       ...a11yNotificationMessages
     };
-    this.a11yNotificationMessages = a11yNotificationMessages;
 
     // Remember the state this application is initialized into, so that we can
     // reset to it later.
@@ -165,8 +164,7 @@ export default class SearchDriver {
     // to the correct default values for the initial UI render
     this.state = {
       ...this.state,
-      ...searchParameters,
-      ...{ a11yNotificationMessages }
+      ...searchParameters
     };
 
     // We'll trigger an initial search if initial parameters contain
@@ -268,9 +266,8 @@ export default class SearchDriver {
               ? totalResults
               : start + resultsPerPage - 1;
 
-          const args = { start, end, totalResults, searchTerm };
-          const message = this.a11yNotificationMessages.searchResults(args);
-          this.actions.a11yNotify(message);
+          const messageArgs = { start, end, totalResults, searchTerm };
+          this.actions.a11yNotify("searchResults", messageArgs);
         }
 
         if (!skipPushToUrl && this.trackUrlState) {
--- a/packages/search-ui/src/SearchDriver.js
+++ b/packages/search-ui/src/SearchDriver.js
@@ -257,6 +257,22 @@ export default class SearchDriver {
           wasSearched: true
         });
 
+        if (this.a11yNotifications) {
+          const { totalResults } = resultState;
+
+          // This is repeated from PagingInfoContainer
+          const start =
+            totalResults === 0 ? 0 : (current - 1) * resultsPerPage + 1;
+          const end =
+            totalResults <= start + resultsPerPage
+              ? totalResults
+              : start + resultsPerPage - 1;
+
+          const args = { start, end, totalResults, searchTerm };
+          const message = this.a11yNotificationMessages.searchResults(args);
+          this.actions.a11yNotify(message);
+        }
+
         if (!skipPushToUrl && this.trackUrlState) {
           // We debounce here so that we don't get a lot of intermediary
           // URL state if someone is updating a UI really fast, like typing
--- a/packages/search-ui/src/SearchDriver.js
+++ b/packages/search-ui/src/SearchDriver.js
@@ -6,6 +6,8 @@ import DebounceManager from "./DebounceManager";
 import * as actions from "./actions";
 import Events from "./Events";
 
+import * as a11y from "./A11yNotifications";
+
 function filterSearchParameters({
   current,
   filters,
@@ -86,7 +88,9 @@ export default class SearchDriver {
     onAutocompleteResultClick,
     searchQuery = {},
     trackUrlState = true,
-    urlPushDebounceLength = 500
+    urlPushDebounceLength = 500,
+    a11yNotifications = true,
+    a11yNotificationMessages = {}
   }) {
     this.actions = Object.entries(actions).reduce(
       (acc, [actionName, action]) => {
@@ -130,6 +134,16 @@ export default class SearchDriver {
       urlState = {};
     }
 
+    // Manage screen reader accessible notifications
+    this.a11yNotifications = a11yNotifications;
+    if (this.a11yNotifications) a11y.getLiveRegion();
+
+    a11yNotificationMessages = {
+      ...a11y.defaultMessages,
+      ...a11yNotificationMessages
+    };
+    this.a11yNotificationMessages = a11yNotificationMessages;
+
     // Remember the state this application is initialized into, so that we can
     // reset to it later.
     this.startingState = {
@@ -151,7 +165,8 @@ export default class SearchDriver {
     // to the correct default values for the initial UI render
     this.state = {
       ...this.state,
-      ...searchParameters
+      ...searchParameters,
+      ...{ a11yNotificationMessages }
     };
 
     // We'll trigger an initial search if initial parameters contain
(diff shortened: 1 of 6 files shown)


constancecchen:
@JasonStoltz and @yakhinvadim - I'd super appreciate your sanity checks on what direction to go with announcing search results.

1. 487e24d - Use `PagingInfoContainer` from `react-search-ui` - this is pretty straightforward but it does mean that non-React users no longer gain 

2. 4e87abf - Use `SearchDriver` from `search-ui` itself - this guarantees that all users will get this screen reader messaging. I think this is preferable, however, as you can see I now have to [repeat the start/end calculation logic from PagingInfoContainer](https://github.com/elastic/search-ui/pull/355/files#diff-bdd309e6d13fa10219eac8138f13fe5dR263).

I'm pretty sure 2. is the way to go - if so, would you all be OK with moving the `start`/`end` logic to SearchDriver.js and passing it down from there for DRYness instead?

packages/search-ui/src/SearchDriver.js
```diff
@@ -151,7 +165,8 @@ export default class SearchDriver {
     // to the correct default values for the initial UI render
     this.state = {
       ...this.state,
-      ...searchParameters
+      ...searchParameters,
+      ...{ a11yNotificationMessages }
```
constancecchen: actually - I'm not 100% sure, but should this really be in state? Or does it feel more like an action?

In theory you are calling functions, e.g. `a11yNotificationMessages.searchResults({ ... })` which makes me think this belongs more in `actions` than in state.

yakhinvadim:
I agree that 2 is the way to go. I'm also for storing the start/end logic in SearchDriver, I can't think of a case where a user would want to calculate it another way. This would also be very convenient for non-React users.

packages/search-ui/src/SearchDriver.js
yakhinvadim: Please, take the following with a grain of salt, I'm not sure if I'm right or not.

They don't look like state - they don't change after initialization.
They also don't look like actions - you don't call them directly to change state.
Aren't they more like a part of a config?

constancecchen: Yeah, I think I actually agree with you. I don't know how else to pass it down to the `withSearch` HOC though - since state and action are what get passed down, and we need views/users to be able to access the message functions somehow.

Should we maybe pass down a whole new key like `getMessages` or `driver.messages`?... 😬

yakhinvadim: > we need views/users to be able to access the message functions somehow
Why? I thought users would add these functions to config once and then not touch it anywhere in the components (in most cases). If (in rare cases) users need to change the a11y message for some reason manually, they would call one universal action a11yNotify.

JasonStoltz: Yes, that was my impression too. I thought the only thing we were going to do with them is have the SearchDriver write them out to a DOM element.

JasonStoltz:
@constancecchen I think you are spot on with option 2. That would be great.

packages/search-ui/src/SearchDriver.js
constancecchen: My impression was that we needed views to have access to the message functions so they can be called within views. I particularly need this not for the "search results" announcement, which can be done in SearchDriver.js but for the "more filters" announcement.

"More Filters" logic is handled entirely within `react-search-ui`'s Facet.js and is almost entirely a view-related concern. Folks rolling their own Vue or non-React components will not use our implementation and therefore the announcement doesn't belong in the top level `search-ui` package.

However, we still need it to support custom messages passed to `a11yNotificationMessages`, and therefore the component needs access to said obj of message functions - presumably through the HOC.

Does that make use case make sense to you both?

JasonStoltz: @constancecchen It does, let me post an example of how I would handle this.

JasonStoltz: This is my suggestion:

SearchDriver:
- Add an action called "a11yNotification" (or something similar) that accepts a string to be read in notifications. i.e., `allyNotification("More was clicked, showing whatever")`

Facet.js
https://github.com/elastic/search-ui/blob/master/packages/react-search-ui/src/containers/Facet.js#L48
- Call `allyNotification` with a message on `handleClickMore`.

How does that sound?

JasonStoltz: If we felt we need to keep all messages together, `allyNotification` could take a string that corresponds to a key in `a11yNotificationMessages`, so instead of reading whatever is passed literally, it would look up a message.

constancecchen: @JasonStoltz It does, and that's what I've done in https://github.com/elastic/search-ui/pull/355/commits/114921e07dbdc86521e0e605994f1fd9d1fce5e1 - the issue is that I still need access to `a11yNotificationMessages` (the message functions) which means passing it down via either state or actions.

The original impetus behind `a11yNotificationMessages` was to let devs customize the screen reader message (for localization, etc.). If I don't use it, I'm passing in a hard coded string and negating that customization.

constancecchen: Hm - the only problem I see with the string/key lookup is that (also per the original [issue writeup](https://github.com/elastic/search-ui/issues/325)) - messages aren't just a static string, we're passing in arguments in order to compose a string, and some of these args are only available in the view itself.

Unless I'm missing some way of both looking up a function and passing in args at the same time? 😕

JasonStoltz: @constancecchen I don't think you should / need to be working with `a11yNotificationMessages` in that component.

```
       const message = this.props.a11yNotificationMessages.moreFilters({
        optionsCount,
        showingAll
      });
      this.props.a11yNotify(message);
```

Could just be:

```
      this.props.a11yNotify("moreFilters", { optionsCount, showingAll});
```

Which is saying "read the moreFilter message, here are the corresponding options".

constancecchen: Ahhh! 🤦‍♀ Yep, that would do it. I don't know why passing more than one param didn't occur to me sooner, thanks so much Jason!!

constancecchen: This makes [the `a11yNotify` action feel much more robust](https://github.com/elastic/search-ui/pull/355/commits/78ffd85d6edc1efab82d9012a40bc5f4d777ab47) now too. Thanks a billion Jason! ✨

constancecchen: Also, just one more follow-up to this refactor that I only just considered - if users want to roll their own completely custom a11y notification for, e.g. their completely custom component - they should accomplish this by extending the `a11yNotificationMessages` obj, correct? So, for example:

```js
SearchDriver({
  a11yNotificationMessages: {
    customMessage: () => 'Hello world',
  },
});
```

This means that:

1. We're now limiting the `a11yNotify` action to not simply allow strings - originally I was thinking users could do something like `a11yNotify('Whatever I want!')`, but that will no longer be the case with this architecture. Are we good with it? (I'm good with it personally, just checking that we're all on the same page.)

2. Per 78ffd85, we're now limiting `a11yNotificationMessages` to only contain functions. For example, if someone tried to pass `customMessage: 'Hello world'`, a type error would get thrown. Are we okay with this restriction? Or do we want to add a `typeof function` check to allow both primitive strings _and_ functions? 🤔 (Feels a bit overkill to me and I'm fine with the limitation, but also wanted to check here that we all agree.)

@JasonStoltz @yakhinvadim - would love your thoughts! 😀

packages/search-ui/src/SearchDriver.js
```diff
@@ -235,13 +248,29 @@ export default class SearchDriver {
...
         this.requestSequencer.completed(requestId);
 
+        // Results/Paging start & end
+        const { totalResults } = resultState;
+        const start =
+          totalResults === 0 ? 0 : (current - 1) * resultsPerPage + 1;
+        const end =
+          totalResults <= start + resultsPerPage
+            ? totalResults
+            : start + resultsPerPage - 1;
```
constancecchen: Following up here on our decision to move `start`/`end` logic out of `PagingInfoContainer` and into `SearchDriver` - are you guys good with this? Or is there a more elegant way/place to put this that I'm missing? 🤔 

Here's the full commit for context: https://github.com/elastic/search-ui/pull/355/commits/4c58a7381f21174a205f61e1ae51a31744e65a57

JasonStoltz: @constancecchen This looks correct to me. Since these have moved to the global state, I think these should be renamed to give some more context. Maybe `pagingStart` and `pagingEnd`?

packages/search-ui/src/SearchDriver.js
JasonStoltz: @constancecchen Probably clearer for users to just always use functions. Easier to document, less code, smaller API surface.

yakhinvadim: (2) I agree with Jason

yakhinvadim: (1) I'm not 100% sure about this, more like 70%, but if you both think that that's ok, I'm in.

constancecchen:
Thanks so much for all the comments and help everyone! Sounds like we're good to go with this direction. I'm going to close this draft PR and open a new one that's been cleaned up and rebased (so we can keep this architecture discussion for reference but have a fresh start for code review). 🙏

PR #356, opened 2019-07-18
Added branching strategy to CONTRIBUTING

JasonStoltz:


CONTRIBUTING.md
```diff
@@ -74,40 +74,57 @@ npm run test -- --watch
 
 The [sandbox](examples/sandbox/README.md) app can be used as a local development aid.
 
+### Branching Strategy
+
+Our branching strategy is based on [Github Flow](https://guides.github.com/introduction/flow/).
```
constancecchen: To be honest, I'm not super sure how helpful this link is? It seems to describe GitHub 101 more than a git branching strategy like gitflow or oneflow for example. I think we could probably get away with just removing this paragraph altogether and starting with "Our master branch..." instead.

JasonStoltz: I agree.

CONTRIBUTING.md
```diff
@@ -74,40 +74,55 @@ npm run test -- --watch
...
+
+Our `master` branch holds the latest development code for the next release. If the next release will be a minor release, the expecation is that no breaking changes will be in `master`. If a change would be breaking, we need to put it behind a feature flag, or make it an opt-in change. We will only merge breaking PRs when we are ready to start working on the next major.
+
+All PRs should be created from a fork, to keep a clean set of branches on `origin`.
+
+Releases should be performed directly in master, following the Publishing guide below.
+
+We use a `stable` branch to indicate the latest release code.
+
+We will create branches for minor releases on if a patch is required. I.e., if a patch is required for `1.1.0`, we would create a `1.1` branch to track patches for the `1.1` release.
```
yakhinvadim: I think `on` is not needed here?

CONTRIBUTING.md
```diff
@@ -74,40 +74,55 @@ npm run test -- --watch
...
-3. Verify correct tags and commits have been created.
-4. Run `nvm use` to make sure you are running the correct version of node, and verify that `npm run build` runs without error before publishing.
-5. Run `npx lerna publish --force-publish=* from-package`.
-6. Verify `0.6.0` has been published to npm.
-7. Create new version branch `git checkout -b v0.6` and push to `origin`.
-8. If this is the latest version of the library, update the `stable` branch to this version `git rebase v0.6` and force push to `origin`.
-9. Create release in Github.
-
-Publish patch version
+2. Commit these changes.
```
yakhinvadim: I think markdown supports putting all the numbers in an ordered list as `1`. It then calculate the right number in place. It helps keeping the git diff clean if you add an item in the middle.

yakhinvadim: Please ignore this, there's no point in doing this in this PR

JasonStoltz: Ah, good point, that's good to know for the future.

JasonStoltz:
@yakhinvadim @constancecchen Great, I think one of you should go through the publish process the next time we have a release.

CONTRIBUTING.md
JasonStoltz: Oops, that should have read:

> We will create branches for minor releases **only** if a patch is required

Issue #337, opened 2019-07-15
Add "" to PagingInfo component


yakhinvadim:
Now if a user doesn't provide a search term it looks a little weird:
![image](https://user-images.githubusercontent.com/11838280/61223096-60552500-a6f2-11e9-9e32-6c6ca29e89f6.png)

It would look better with "":
![image](https://user-images.githubusercontent.com/11838280/61223116-69de8d00-a6f2-11e9-87be-b91d71bbf0ed.png)

constancecchen:
Alternatively, would it also make sense to get rid of the `for` if we're purposely doing a blanket filter/category search term? e.g.,

`Showing 1-20 out of 2005:`

Kacppian:
@constancecchen  I agree with you, removing "for" makes more sense. I just tried out both of them, I like removing the "for". Here's some screenshots - 

**Without `for`**
![Screenshot from 2019-07-20 01-25-06](https://user-images.githubusercontent.com/14990078/61562013-f63fd580-aa8d-11e9-855e-3cb7a577eaca.png)

**With quotes**
![Screenshot from 2019-07-20 01-25-28](https://user-images.githubusercontent.com/14990078/61562148-43bc4280-aa8e-11e9-9911-e40176261622.png)

JasonStoltz:
@Kacppian I agree, removing the for entirely looks much better.

Kacppian:
Awesome, I'll implement the change then. 

I see `lib/` `es/` and `src/` in the directories under `react-search-ui-views`. Could you please help me understand why that is the case? I'm a little lost. There are a bunch of PagingInfo.js files, do I have to make the same change across the files?

Kacppian:
Do we remove just the word `for` and still show the `searchTerm` or we want to get rid of everything that follows `for`?

Kacppian:
Submitted the PR

JasonStoltz:
Closed by https://github.com/elastic/search-ui/pull/360

PR #360, opened 2019-07-22
Modifies the way empty string searches are displayed in the Pagination info

Kacppian:
## Description

When a user searches for an empty string, the UI showing the pagination information looks a little odd. It looks much better when we remove the text "for". 

## List of changes
* Removes "for" from the text displaying the pagination information. 

## Associated Github Issues
#337

### DIFF
--- a/packages/react-search-ui-views/src/PagingInfo.js
+++ b/packages/react-search-ui-views/src/PagingInfo.js
@@ -3,14 +3,14 @@ import React from "react";
 
 import { appendClassName } from "./view-helpers";
 
-function PagingInfo({ className, end, start, totalResults }) {
+function PagingInfo({ className, end, searchTerm, start, totalResults }) {
   return (
     <div className={appendClassName("sui-paging-info", className)}>
       Showing{" "}
       <strong>
         {start} - {end}
       </strong>{" "}
-      out of <strong>{totalResults}</strong>
+      out of <strong>{totalResults}</strong>: <em>{searchTerm}</em>
     </div>
   );
 }
--- a/packages/react-search-ui-views/src/PagingInfo.js
+++ b/packages/react-search-ui-views/src/PagingInfo.js
@@ -10,7 +10,12 @@ function PagingInfo({ className, end, searchTerm, start, totalResults }) {
       <strong>
         {start} - {end}
       </strong>{" "}
-      out of <strong>{totalResults}</strong>: <em>{searchTerm}</em>
+      out of <strong>{totalResults}</strong>{" "}
+      {searchTerm && (
+        <React.Fragment>
+          for: <em>{searchTerm}</em>
+        </React.Fragment>
+      )}
     </div>
   );
 }
--- a/packages/react-search-ui-views/src/PagingInfo.js
+++ b/packages/react-search-ui-views/src/PagingInfo.js
@@ -10,9 +10,10 @@ function PagingInfo({ className, end, searchTerm, start, totalResults }) {
       <strong>
         {start} - {end}
       </strong>{" "}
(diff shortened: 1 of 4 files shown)


packages/react-search-ui-views/src/PagingInfo.js
```diff
@@ -10,7 +10,7 @@ function PagingInfo({ className, end, searchTerm, start, totalResults }) {
       <strong>
         {start} - {end}
       </strong>{" "}
-      out of <strong>{totalResults}</strong> for: <em>{searchTerm}</em>
```
constancecchen: I don't think we should delete the `for` word completely - only if `searchTerm` is missing. So we should really be wrapping this in a conditional. Does that make sense?

So if a `searchTerm` is present, the string should still be:
```Showing X - Y out of Z for: FooBar```
and if not:
```Showing X - Y out of Z:```

@JasonStoltz / @yakhinvadim Feel free to let me know if you disagree and I'll update my review :)

Comment (JasonStoltz) [reply packages/react-search-ui-views/src/PagingInfo.js]: Yes but one small correction...

So if a searchTerm is present, the string should still be:

```
Showing X - Y out of Z for: FooBar
```

and if not:

```
Showing X - Y out of Z
```

No semi-colon in the second case, right?

Comment (constancecchen) [reply packages/react-search-ui-views/src/PagingInfo.js]: Oh yeah, no colon. I'll update my comment!

Comment (Kacppian) [reply packages/react-search-ui-views/src/PagingInfo.js]: Cool, I'll make the required changes

Comment (Kacppian) [reply packages/react-search-ui-views/src/PagingInfo.js]: @JasonStoltz @constancecchen Pushed some commits that address this. Do let me know if you'd like it to be done in a different way.

File: packages/react-search-ui-views/src/PagingInfo.js [hunk 2]
```diff
@@ -10,7 +10,12 @@ function PagingInfo({ className, end, searchTerm, start, totalResults }) {
...
+      out of <strong>{totalResults}</strong>{" "}
+      {searchTerm && (
+        <React.Fragment>
```
constancecchen: You can use the [fragment shorthand](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) `<>...</>` here instead of `<React.Fragment>...</React.Fragment>` - it's what we've been using elsewhere in the repo :)

packages/react-search-ui-views/src/PagingInfo.js
```diff
@@ -10,7 +10,12 @@ function PagingInfo({ className, end, searchTerm, start, totalResults }) {
       <strong>
         {start} - {end}
       </strong>{" "}
-      out of <strong>{totalResults}</strong> for: <em>{searchTerm}</em>
+      out of <strong>{totalResults}</strong>{" "}
```
constancecchen: Is it possible to move the `{" "}` to line 16 instead (before the '`for`)? Ideally the extra space should only show up if needed if the word "for" appears. If Prettier gives you issue with that though, then no worries.

constancecchen:
Just some super minor comments - the main logic looks great!

packages/react-search-ui-views/src/PagingInfo.js
Kacppian: I've made this change, prettier didn't throw any errors. I should've thought about this. Thanks for pointing it out ~!

packages/react-search-ui-views/src/PagingInfo.js
Kacppian: I just noticed after you'd mentioned. Made the change for it :)

PR #366, opened 2019-07-24
[WIP, Discussion] Documentation changes for 1.1

goodroot:
## Description

Adds responsive to features and documents screen reading.

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -7,6 +7,7 @@
 5. [Build Your Own Component](#build-your-own-component)
 6. [Connectors and Handlers](#connectors-and-handlers)
 7. [Performance](#performance)
+8. [Accessibility](#accessibility)
 
 # Headless Core
 
@@ -1392,3 +1393,16 @@ this.setState(prevState => ({ options: [...prevState.options, "newOption"] }));
 If you ever need to debug performance related issues, see the instructions in the Optimizing Performance guide for
 enabling the "Highlight Updates" feature in the
 [React Developer tools for Chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi).
+
+# Accessibility
+
+Search UI supports screen readers.
+
+Place `hasA11yNotifications: true` within the config to enable:
+
+```jsx
+config={{
+  apiConnector: connector,
+  hasA11yNotifications: true
+}}
+```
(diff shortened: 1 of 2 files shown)


ADVANCED.md
```diff
@@ -1392,3 +1393,16 @@ this.setState(prevState => ({ options: [...prevState.options, "newOption"] }));
 If you ever need to debug performance related issues, see the instructions in the Optimizing Performance guide for
 enabling the "Highlight Updates" feature in the
 [React Developer tools for Chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi).
+
+# Accessibility
```
constancecchen: I might actually suggest not including an entire accessibility section for now (at least until 2.0, when we ship with `hasA11yNotifications` that defaults to `true`)?

I'm also not 100% done with all the accessibility auditing and testing in the app (for example, the SingleLinksFacet component still isn't accessible, and our pagination is a bit janky for screen reader usage), so I'd feel a little silly making the announcement before 2.0/before I feel confident saying Search UI is 100% accessible.

constancecchen: I'd also be thrilled if accessibility got its own page (maybe when we move to a full-fledged doc site?). There's a couple other accessibility features I'd like to highlight, such as, for example, being able to localize/customize screen reader announcements, or being able to roll your own announcement. That would look something like this:

```js
config={{
  apiConnector: connector,
  a11yNotificationMessages: {
    searchResults: ({ start, end, totalResults, searchTerm}) => {
      // Example localization
      return `Mostrando ${start}-${end} de ${totalResults}, buscando "${searchTerm}"`;
    }
  }
}}
```

And rolling your own custom screen reader notification would look something like this:

```js
config={{
  apiConnector: connector,
  a11yNotificationMessages: {
    customMessage: () => {
      return 'Hello world!'
    }
  }
}}
```
```jsx
const CustomComponent = (props) => {
  const onClick = props.a11yNotify('customMessage');

  return (
    <button type="button" onClick={onClick}>Hello world!</button>
  );
}

export default withSearch(({ a11yNotify }) => ({ a11yNotify }))(CustomComponent);
```

README.md
```diff
@@ -25,7 +25,7 @@ search experience up and running in minutes.
 ### Features :+1:
 
 - **You know, for search** - Maintained by [Elastic](https://elastic.co), the team behind Elasticsearch.
-- **Speedy Implementation** - Build a complete search experience with a few lines of code.
+- **Speedy Implementation** - Build a complete, responsive search experience with a few lines of code.
```
constancecchen: See above comment - if we want to keep it simple for now, I'm happy to lose the accessibility section for now and have this be `Build a complete, responsive, and accessible search experience with a few lines of code` :)

Issue #241, opened 2019-05-21
Each component should accept all html props


JasonStoltz:
https://reakit.io/docs/styling/

Accepting all HTML props makes it much easier to customize the components in the library.

Example:

In addition to the required and known parameters that PagingInfo accepts, it should also pass through any other prop that it receives:

```
<PagingInfo style={{display: "flex"}} />
```

To make this change, in the `PagingInfo.js` container, you need to make sure to pass through the additional props to the corresponding view:

`react-search-ui - PagingInfo.js`
```
const {
      className,
      current,
      resultsPerPage,
      resultSearchTerm,
      totalResults,
      view,
      ...rest
    } = this.props;

return View({
      className,
      end: end,
      searchTerm: resultSearchTerm,
      start: start,
      totalResults: totalResults,
      ...rest
    });
```

In the corresponding view, you would render those additional props to the root element:

`react-search-ui-views - PagingInfo.js`
```
function PagingInfo({ className, end, searchTerm, start, totalResults, ...rest }) {
  return (
    <div className={appendClassName("sui-paging-info", className)} {...rest}>

    </div>
  );
}
```

Jfelix61:
I can do this.

JasonStoltz:
@Jfelix61 Great! Feel free to work on this and submit a PR, and let us know if you have any questions.

Jfelix61:
Submitted the PR. However circle-ci tests fail since the SearchBox component receives new props: result of the rest implementation.

JasonStoltz:
Closed by #368

PR #368, opened 2019-07-25
Enhancement/rest props

Jfelix61:
## Description
As per title. Since SearchBox now receives new props from ...rest, test snapshot had to be updated.
## List of changes
-Snapshot of SearchBox 
-Components now receive all html props 
## Associated Github Issues
#241

### DIFF
--- a/packages/react-search-ui/src/containers/PagingInfo.js
+++ b/packages/react-search-ui/src/containers/PagingInfo.js
@@ -22,7 +22,8 @@ export class PagingInfoContainer extends Component {
       pagingEnd,
       resultSearchTerm,
       totalResults,
-      view
+      view,
+      ...rest
     } = this.props;
 
     const View = view || PagingInfo;
@@ -32,7 +33,8 @@ export class PagingInfoContainer extends Component {
       searchTerm: resultSearchTerm,
       start: pagingStart,
       end: pagingEnd,
-      totalResults: totalResults
+      totalResults: totalResults,
+      ...rest
     });
   }
 }
--- a/packages/react-search-ui/src/containers/PagingInfo.js
+++ b/packages/react-search-ui/src/containers/PagingInfo.js
@@ -14,7 +14,7 @@ export class PagingInfoContainer extends Component {
     pagingEnd: PropTypes.number.isRequired,
     resultSearchTerm: PropTypes.string.isRequired,
     totalResults: PropTypes.number.isRequired,
-    results: PropTypes.arrayOf(ResultType).isRequired
+    results: PropTypes.arrayOf(ResultType)
   };
 
   render() {
--- a/packages/react-search-ui/src/containers/PagingInfo.js
+++ b/packages/react-search-ui/src/containers/PagingInfo.js
@@ -2,7 +2,6 @@ import PropTypes from "prop-types";
 import { Component } from "react";
 import { withSearch } from "..";
 import { PagingInfo } from "@elastic/react-search-ui-views";
-import { Result as ResultType } from "../types";
 
 export class PagingInfoContainer extends Component {
   static propTypes = {
@@ -13,8 +12,7 @@ export class PagingInfoContainer extends Component {
     pagingStart: PropTypes.number.isRequired,
     pagingEnd: PropTypes.number.isRequired,
     resultSearchTerm: PropTypes.string.isRequired,
-    totalResults: PropTypes.number.isRequired,
-    results: PropTypes.arrayOf(ResultType)
+    totalResults: PropTypes.number.isRequired
   };
 
   render() {
@@ -24,8 +22,6 @@ export class PagingInfoContainer extends Component {
       pagingEnd,
       resultSearchTerm,
       totalResults,
-      // eslint-disable-next-line no-unused-vars
-      results,
       view,
       ...rest
     } = this.props;
--- a/packages/react-search-ui/src/containers/PagingInfo.js
+++ b/packages/react-search-ui/src/containers/PagingInfo.js
@@ -2,6 +2,7 @@ import PropTypes from "prop-types";
 import { Component } from "react";
 import { withSearch } from "..";
 import { PagingInfo } from "@elastic/react-search-ui-views";
+import { Result as ResultType } from "../types";
 
 export class PagingInfoContainer extends Component {
   static propTypes = {
@@ -12,7 +13,8 @@ export class PagingInfoContainer extends Component {
     pagingStart: PropTypes.number.isRequired,
     pagingEnd: PropTypes.number.isRequired,
     resultSearchTerm: PropTypes.string.isRequired,
-    totalResults: PropTypes.number.isRequired
+    totalResults: PropTypes.number.isRequired,
+    results: PropTypes.arrayOf(ResultType).isRequired
   };
 
   render() {
@@ -22,6 +24,8 @@ export class PagingInfoContainer extends Component {
       pagingEnd,
       resultSearchTerm,
       totalResults,
+      // eslint-disable-next-line no-unused-vars
+      results,
       view,
       ...rest
     } = this.props;
--- a/packages/react-search-ui/src/containers/PagingInfo.js
+++ b/packages/react-search-ui/src/containers/PagingInfo.js
@@ -2,6 +2,7 @@ import PropTypes from "prop-types";
 import { Component } from "react";
 import { withSearch } from "..";
 import { PagingInfo } from "@elastic/react-search-ui-views";
+import { Result as ResultType } from "../types";
 
 export class PagingInfoContainer extends Component {
   static propTypes = {
@@ -12,7 +13,8 @@ export class PagingInfoContainer extends Component {
     pagingStart: PropTypes.number.isRequired,
     pagingEnd: PropTypes.number.isRequired,
     resultSearchTerm: PropTypes.string.isRequired,
-    totalResults: PropTypes.number.isRequired
+    totalResults: PropTypes.number.isRequired,
+    results: PropTypes.arrayOf(ResultType)
   };
 
   render() {
@@ -22,6 +24,8 @@ export class PagingInfoContainer extends Component {
       pagingEnd,
       resultSearchTerm,
       totalResults,
+      // eslint-disable-next-line no-unused-vars
+      results,
       view,
       ...rest
     } = this.props;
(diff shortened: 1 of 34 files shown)


constancecchen:
Could you remove/revert the `package-lock.json` file changes from this PR since it's not directly relevant to the issue? That way if we need to roll back this PR for any reason we don't also roll back unrelated changes or changes that we do want.

It's very possible that we forgot to commit package-log changes at some point and do need to update said files, but let's do that in a separate chore PR just imo (CC @JasonStoltz / @yakhinvadim).

Jfelix61:
Done accordingly

JasonStoltz:
@constancecchen Yeah I agree. Related, I think for the individual packages we should just remove the lock files: https://github.com/elastic/enterprise-search-node/issues/8.

JasonStoltz:
It looks like we're actually passing too many props through to the views:

React_App

We are using `...rest` correctly here: https://github.com/elastic/search-ui/blob/3fe8d92016aeb0d26921ebd19f4a0cc4a0f48962/packages/react-search-ui/src/containers/Result.js#L46, however, this is not accounting for all of the components props. Like `shouldTrackClickThrough` for instance: https://github.com/elastic/search-ui/blob/3fe8d92016aeb0d26921ebd19f4a0cc4a0f48962/packages/react-search-ui/src/containers/Result.js#L34. It IS being used, but it's not being destructured in the render function, so it gets included in `rest`.

This would not be an issue we were to convert the to function components, which is something we should think about doing sooner rather than later.

Anyone know of a good way to handle this? One thought, not sure how great of a thought it is though, would be to calculate the "additionalProps" / "rest" by filtering out all of the props specified by keys specified `propTypes`.

So something like
```
function getAdditinonalProps(props) {
  const propsToFilterOut = Object.keys(ResultContainer.propTypes);
  return props.reduce(/* Some function that filters out props based on propsToFilterOut */)
}
```

JasonStoltz:
Here's what the React docs have to say about it, btw: https://zhenyong.github.io/react/docs/transferring-props.html#transferring-with-...-in-jsx

JasonStoltz:
The other thing we could do is just make it a pattern to *always* access all props through the pass them through as needed.

https://github.com/elastic/search-ui/blob/3fe8d92016aeb0d26921ebd19f4a0cc4a0f48962/packages/react-search-ui/src/containers/Result.js#L54

So something more like:

```
 function  handleClickLink(clickThroughTags, shouldTrackClickThrough, trackClickThrough, id) {
    if (shouldTrackClickThrough) {
      trackClickThrough(id, clickThroughTags);
    }
  };
```
```
onClickLink: () => this.handleClickLink(clickThroughTags, shouldTrackClickThrough, trackClickThrough, result.id.raw)
```

If we did this, then `...rest` would always be accurate, and this style actually gets us a bit closer to  function components. 

I'm not sure if that would work in all cases or not.

Jfelix61:
I believe converting it to a functional component would be the best idea, since it is something that has to be done, regardless of the extra work. Which library should we focus if we were to choose to restructure class components? 

Regarding the filtering of props, wouldn't it result in less performance since we would be filtering on each component with rest operator, meaning a O(n) time complexity?

yakhinvadim:
> Anyone know of a good way to handle this (besides with a function component)?

Before I simply excluded props that I didn't want to pass down. Like this:
```
const {
      className,
      result,
      titleField,
      urlField,
      view,
+     clickThroughTags,
+     shouldTrackClickThrough,
+     trackClickThrough
      ...rest
    } = this.props;
```

Although a bit manual, I see it as explicit and precise. This shouldn't be too hard to add since from the console screen above it looks like we have only 3 such warnings.

I guess rewriting the components with functions would be the right solution here, but this seems like too big of a change now.

For this PR I would go with the manual exclusion.

yakhinvadim:
@Jfelix61 Do you see any tests failing locally? All the failing tests in CI are not matching snapshots.

JasonStoltz:
@yakhinvadim I agree Vadim, I think a simple and concise fix here might be best.

Also, the specs are failing due to this issue, I believe. The additional props are being passed down.

yakhinvadim:
@Jfelix61 It would be great to have tests that verify that additional props passed through to the views. They would help us be sure that we didn't lose that feature when we rewrite the class components with functions. 

There are tests named "passes className through to the view". Example: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui/src/containers/__tests__/Paging.test.js#L49-L60
I guess we can test the changes in this PR following the same pattern: confirming that some arbitrary prop is passed along instead of 'className'.

Jfelix61:
> @Jfelix61 Do you see any tests failing locally? All the failing tests in CI are not matching snapshots.

Yes.

> > Anyone know of a good way to handle this (besides with a function component)?
> 
> Before I simply excluded props that I didn't want to pass down. Like this:
> 
> ```
> const {
>       className,
>       result,
>       titleField,
>       urlField,
>       view,
> +     clickThroughTags,
> +     shouldTrackClickThrough,
> +     trackClickThrough
>       ...rest
>     } = this.props;
> ```
> 
> Although a bit manual, I see it as explicit and precise. This shouldn't be too hard to add since from the console screen above it looks like we have only 3 such warnings.
> 
> I guess rewriting the components with functions would be the right solution here, but this seems like too big of a change now.
> 
> For this PR I would go with the manual exclusion.

I find it as a really good alternative. 

> @Jfelix61 It would be great to have tests that verify that additional props passed through to the views. They would help us be sure that we didn't lose that feature when we rewrite the class components with functions.
> 
> There are tests named "passes className through to the view". Example: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui/src/containers/__tests__/Paging.test.js#L49-L60
> I guess we can test the changes in this PR following the same pattern: confirming that some arbitrary prop is passed along instead of 'className'.

I would gladly add them. However, which props should I be checking for?

yakhinvadim:
> I would gladly add them. However, which props should I be checking for?

Great. I think testing that the 'style' property is passed down could be useful. It is widely used and is valid for almost all html elements. I'm open to better suggestions though.

constancecchen:
Awesome catches all - strongly agreed that for now, let's explicitly pull out non-valid attributes and save refactoring for another day/as tech debt.

For writing tests - I'd suggest a `data-foo="bar"` prop/attribute of some kind. Data attributes are pretty safe to test as they're fairly benign as well as being a common use-case (e.g. for Cypress testing). Another option would be `id`, as that's another commonly used attribute.

yakhinvadim:
Great idea, Constance. Let's use data-foo for testing.

Jfelix61:
Added all accordingly, had to disable eslint for unused props and implemented to PropTypes the props explicitly added.

Jfelix61:
> Added all accordingly, had to disable eslint for unused props and implemented to PropTypes to the props explicitly added.

Jfelix61:
I'm trying `npm run watch` at root directory but getting errors that ./bin/watch-js is not found. However sanbox runs without console error. 

image

In the meantime I'll be deleting style test.

Jfelix61:
Can't seem to `npm run watch` on the root folder, seem to be getting watch-js not found error. Also sanbox can't compile since it can't find `@elastic/react-search-ui-views/lib/styles/styles.css`.  However, in the meantime, will be deleting style test.

constancecchen:
Try re-running `npm install` from root? Unfortunately dependencies can get a bit tricky since we're using Lerna, so worst comes to worst you can nuke your local folder and re-pull down from remote/GitHub and re-install.

Also, you can do `npm run sandbox` from the root folder instead of manually `cd`ing into sandbox as a heads up (this won't fix your issues above, just mentioning it as a nice shortcut).

elasticcla:
Hi @Jfelix61, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/search-ui/pull/368.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

packages/react-search-ui/src/containers/Sorting.js
```diff
@@ -44,7 +48,12 @@ export class SortingContainer extends Component {
       sortDirection,
       sortField,
       sortOptions,
-      view
+      view,
+      // eslint-disable-next-line no-unused-vars
+      results,
```
yakhinvadim: Sorry if I'm wrong, but why we need to exclude `results` and `searchTerm`  from the `...rest` here?
It looks like for this file we could remove them both from `this.props` destructuring and from propTypes.
Does it make sense?

packages/react-search-ui/src/containers/ResultsPerPage.js
```diff
@@ -12,7 +13,9 @@ export class ResultsPerPageContainer extends Component {
     // State
     resultsPerPage: PropTypes.number.isRequired,
     // Actions
-    setResultsPerPage: PropTypes.func.isRequired
+    setResultsPerPage: PropTypes.func.isRequired,
+    results: PropTypes.arrayOf(ResultType),
```
yakhinvadim: This is similar to Sorting.js container:
> Sorry if I'm wrong, but why we need to exclude results and searchTerm from the ...rest here?
It looks like for this file we could remove them both from this.props destructuring and from propTypes.
Does it make sense?

packages/react-search-ui/src/containers/PagingInfo.js
```diff
@@ -12,7 +13,8 @@ export class PagingInfoContainer extends Component {
     pagingStart: PropTypes.number.isRequired,
     pagingEnd: PropTypes.number.isRequired,
     resultSearchTerm: PropTypes.string.isRequired,
-    totalResults: PropTypes.number.isRequired
+    totalResults: PropTypes.number.isRequired,
+    results: PropTypes.arrayOf(ResultType)
```
yakhinvadim: This is similar to Sorting.js container (but only relates to `results` prop):
> Sorry if I'm wrong, but why we need to exclude results and searchTerm from the ...rest here?
It looks like for this file we could remove them both from this.props destructuring and from propTypes.
Does it make sense?

Jfelix61:
Done accordingly. If there is any change pending I'll be glad to help!

yakhinvadim:
Oops, sorry, I had to refresh the page before posting this 🙂 . Everything is good now, thank you for your contribution, @Jfelix61!

PR #373, opened 2019-08-20
Updated licensing

JasonStoltz:


yakhinvadim:
I think we can delete all the License files except the one in the root. It should cover all the content of a repo.
The React monorepo does it that way: they have one top-level license file https://github.com/facebook/react and no licenses in package folders https://github.com/facebook/react/tree/master/packages/react

JasonStoltz:
@yakhinvadim Oh I actually forgot about those. Actually, I had asked when I originally started this project. At the time, it was recommended to me that I add the LICENSE to every package because that is what actually gets published. If I don't, then no LICENSE file is actually shipped in the published package. Does that make sense?

Issue #372, opened 2019-08-19
Make the default placeholder text more universal


yakhinvadim:
Instead of "Search your documents..." we could have a simple "Search", which is better as a default.

Jfelix61:
Hi yakhinvadim, I would like to do this.

PR #374, opened 2019-08-22
Make the default placeholder text more universal

Jfelix61:
## Description
Changed placeholder to be more universal.
## List of changes
-Changed search placeholder
-Modified test to match changes.
## Associated Github Issues
#372


JasonStoltz:
@yakhinvadim I think we'll need to consider this a breaking change. So we'll need to queue this up for a 2.0 release.

JasonStoltz:
@yakhinvadim Do you agree with that?

yakhinvadim:
Yes, you're right

Issue #65, opened 2019-02-05
Implement method of combining actions


JasonStoltz:
See Debounce usage in elastic-co-search: https://github.com/elastic/elastic-co-search/commit/bb07ff13d756264d83f6a6776c14dcc106ee701d

Two actions should be able to be called one after the other, and both actions should be applied in batch, in just a single API call.

Document this as an Advanced Feature in the README

JasonStoltz:
Related:

https://discuss.elastic.co/t/pass-additional-filters-with-search-text-from-search-ui-to-app-search/199647/6
https://github.com/elastic/search-ui/issues/387

JasonStoltz:
Closed by https://github.com/elastic/search-ui/pull/391

PR #391, opened 2019-09-20
Batch actions and the new "clearFilters" option

JasonStoltz:
## Description

In its current state, Search UI has a problem. Actions trigger API calls to the Search API. When there are in-flight calls to the Search API, Search UI will block any new actions.

That means, if a user were to call multiple actions in a row like this the following, that only the first action would be applied.

```jsx
addFilter("states", "Alaska", "any");
addFilter("states", "California", "any");
setSearchTerm("park");
```

One way to fix this, would be to simply unblock subsequent actions from making API calls. However, you would then encounter an issue where the API is making excessive API calls. If you could imagine in the example above, that would result in 3 separate API calls.

In order to cleanly avoid this, I've added the ability for actions to be automatically batched. 

An action can be thought of in 2 parts:
1. A state update
2. A side-effect

The primary side-effect of most actions is to execute a new query. All I've done is debounced that side effect, so only after the user is done calling a series of actions, does the side effect get executed.

### Why would someone need to do this?

We currently don't have any sort of bulk API for applying various filters. Meaning, if you had a reason to apply 2 filters at once, there is no single action to do that. You must make two calls to the addFilter action.

```jsx
addFilter("states", "Alaska", "any");
addFilter("world_heritage_site", "true");
```

Another case, is a search experience with a submit button, like an "Advanced Search" screen. These experiences typically are not updated live as you are making filter selections. Instead, you enter all of your selections and press a submit button to apply them at once.

In the future, we should have a better solution for this. But for now, that could be accomplished by simply hooking the "Submit" button up to a handler that calls all of the appropriate actions to apply the user's selections.

### The case of "setSearchTerm" and "clearFilters"

Currently, "setSearchTerm" clears out existing filters any time that it is executed. This is usually, but not always, desirable. Especially if you think about the case I mentioned above, where you are implementing an "Advanced Search" feature. Calling items in this order actually ends up clearing the filters that you just added when you call "setSearchTerm":

```jsx
addFilter("states", "Alaska", "any");
addFilter("states", "California", "any");
setSearchTerm("park");
```

I added the ability to work around this with a "clearFilters" flag:

```jsx
addFilter("states", "Alaska", "any");
addFilter("states", "California", "any");
setSearchTerm("park", { clearFilters: false });
```

### The case of "setSearchTerm" and "debounce"

I had to think about what the implications of debouncing both the search side effect and also allowing users to debounce searches independently when using "setSearchTerm". For example:

```
setSearchTerm("park", { refresh: true, debounce: 1000 });
addFilter("states", "Alaska", "any");
addFilter("states", "California", "any");
```

What happened in this case, is that 2 queries were being executed. 
1. The initial query which applied the search term "park" as well as the 2 filters on "states".
2. The second query, on "park", with NO filters applied. This is because the "setSearchTerm" side-effect was deferred by 1000 ms and ended up executing after the initial query.

To overcome this, I added logic that cancels any outstanding side-effects, as they should be considered stale and no longer valid to run if a new side-effect is applied.

### Want to test this locally?

Add the following button to `App.js` in the sandbox example:

```jsx
<button
  onClick={() => {
    addFilter("states", "Alaska", "any");
    setSearchTerm("park", { clearFilters: false });
    addFilter("states", "California", "any");
  }}
>
  Click here
</button>
```

Clicking that button BEFORE my changes will only apply the California filter to results.

AFTER my changes, that will apply all 3 of those filters.

## List of changes

- Added batch actions
- Added a "clearFilters" flag to setSearchTerm
- Added side-effect cancelling in our actions
- Added the "clearFilters" option to the SearchBox component as well.

## Associated Github Issues

https://github.com/elastic/search-ui/issues/65


packages/react-search-ui-views/src/SearchBox.js
```diff
@@ -22,6 +22,8 @@ function SearchBox(props) {
     onSubmit,
     useAutocomplete,
     value,
+    // NOTE: These are explicitly de-structured but no used so that they are
```
constancecchen: Typo nit: `no used` -> `not used`

ADVANCED.md
```diff
@@ -529,6 +543,7 @@ page for suggestions, and maintaining the default behavior when selecting a resu
...
 | className                     | String                                                                       | no        |                                                                    |         |                                                                                                                                                                                                                                                                                                                                                              |
+| clearFilters                  | Boolean                                                                      | no        | true                                                               |         | Should existing filters be cleared when a new search is performed?                                                                                                                                                                                                                                                                                           |
```
constancecchen: Sorry, I think this one confuses me a bit. Just to clarify: we're adding *two* new places to pass `clearFilters` - one within `setSearchTerm` options and one on the `<SearchBox>` component? So usage looks like this:

`setSearchTerm('foo', { clearFilters: true })`
`<SearchBox clearFilters />`

Is that correct?

EDIT: Ah okay sorry, I just read further down in the PR description and I see that described in the changelog now. One thing I might suggest is adding an example section somewhere (either in the ADVANCED.md file or a specific doc for SearchBox, perhaps) what that behavior looks like. Although it also might be better to save that for when we switch to a documentation generator like Styleguidist or Storybook :)

constancecchen: Also, one thing I wanted to quickly note - `clearFilters()` is a function/action elsewhere but a boolean here. To my mind, adding multiple variables with the same name and having them be distinctly different types is pretty confusing. I would personally suggest `shouldClearFilters` instead to make it clear which is a bool and which is a function, but that's just my [preference for naming booleans](https://www.samanthaming.com/tidbits/34-better-boolean-variable-names). :)

packages/react-search-ui/src/containers/__tests__/SearchBox.test.js
```diff
@@ -420,6 +479,25 @@ describe("onSelectAutocomplete", () => {
     expect(passedAutocompleteResults).toBeDefined();
     expect(passedDefaultOnSelectAutocomplete).toBeDefined();
   });
+
+  it("will call setSearchTerm if no onSelectAutocomplete is specified and a suggestion is selected", () => {
```
constancecchen: Is this one related to the new clearFilters logic or just something you happened to notice along the way? I think it's great either way, but my one comment for the file overall is that it's getting pretty long and hard to reason about, so I might suggest either breaking up visually with `/**` comment docblocks or grouping all tests with `describe()` parents (there's a couple loose tests that are kind of floating around). That may be something for another day/as tech debt though :)

ADVANCED.md
JasonStoltz: @constancecchen That is a *great* point. One that you have mentioned before but I clearly forgot. Thank you, I will make this change 👍

JasonStoltz:
@constancecchen Yes, sorry for dragging you into this big PR :( I sincerely appreciate it!

packages/search-ui/src/setupTests.js
```diff
@@ -0,0 +1,3 @@
+// We use fake timers to avoid having to use async programming
+// in our tests.
+jest.useFakeTimers();
```
constancecchen: At first glance, I was a little leery of this being in a global setup file instead of being used per-test or per-file on an as-needed basis. That being said, I can't say exactly why, other than that's just what I've done in the past 🤣

One possible issue from a quick Google is that `useFakeTimers()` appears to [currently cause issues with native Promises](https://github.com/facebook/jest/issues/7151), although I'm not sure if we use those anywhere in the codebase - so if that's not an issue for us, I'm good with this 👍

packages/search-ui/src/__tests__/SearchDriver.test.js
```diff
@@ -89,7 +89,7 @@ it("will default facets to {} in state if facets is missing from the response",
     }
   });
 
-  expect(doesStateHaveResponseData(stateAfterCreation)).toBe(true);
+  //expect(doesStateHaveResponseData(stateAfterCreation)).toBe(true);
```
constancecchen: FYI, looks like this was a comment accidentally left in. Should it be uncommented or entirely removed?

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
+   * referential identity of a function. This works by comparing provided function names.
+   *
+   * This also has the ability to short-circuit a debounce all-together, if no wait
+   * time is provided.
+   *
+   * Assumption:
+   * Functions are debounced on a combination of unique function name and wait times. So debouncing won't work on
+   * subsequent calls with different wait times or different functions. That also means that the debounce manager
+   * can be used for different functions in parallel, and keep the two functions debounced separately.
+   *
```
constancecchen: Amazing docblock comment - really helped me understand the what, why, and how of this manager! 🙏 

One quick lint nit here - not sure if intentional, but there's two newlines between the comment and @params and one can possibly be deleted as an extra.

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
+   *
+   * @param {number} wait Milliseconds to debounce. Executes immediately if falsey.
+   * @param {function} fn Function to debounce
+   * @param {function} functionName Name of function to debounce, used to create a unique key
+   * @param  {...any} parameters Parameters to pass to function
```
constancecchen: Nit: Extra whitespace between `@param` and `{...any}`

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
+   */
+  runWithDebounce(wait, functionName, fn, ...parameters) {
     if (!wait) {
       return fn(...parameters);
     }
 
-    const key = fn.toString() + wait.toString();
+    const key = functionName + wait.toString();
```
constancecchen: Just curious - was `fn.toString()` not working previously? What was the impetus to change this to an explicitly passed functionName string? (It seems rather clever, hence my curiosity)

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
+      .forEach(([_, cachedValue]) => cachedValue.cancel());
+  }
 }
+/**
+ * Perform a standard debounce
+ *
+ * @param {number} wait Milliseconds to debounce. Executes immediately if falsey.
+ * @param {function} fn Function to debounce
+ */
+DebounceManager.debounce = (wait, fn) => {
```
constancecchen: Just curious - why put this outside the class instead of keeping it inside? e.g.

```js
class DebounceManager {
  ...

  /**
   * ...
   */
  debounce(wait, fn) {
    return debounceFn(fn, { wait });
  }
}
```

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
     if (!debounced) {
       this.debounceCache[key] = debounceFn(fn, { wait });
       debounced = this.debounceCache[key];
     }
     debounced(...parameters);
   }
+
+  cancelByName(functionName) {
+    Object.entries(this.debounceCache)
+      .filter(([cachedKey]) => cachedKey.startsWith(functionName))
```
constancecchen: I'm a little leery here of using `startsWith` instead of just `cachedKey === functionName`. It would return a false positive for `foo()` vs. `fooBar()`, however unlikely. Is there a particular reason to use `startsWith`?

packages/search-ui/src/setupTests.js
JasonStoltz: @constancecchen OK, great, I'm glad you're ok with it. It makes writing these tests infinitely easier.

packages/search-ui/src/__tests__/SearchDriver.test.js
JasonStoltz: @constancecchen Whoops, I will take care of that.

packages/react-search-ui/src/containers/__tests__/SearchBox.test.js
JasonStoltz: @constancecchen It's one that I noticed was missing. And I agree, as I was writing this I was pretty overwhelmed. If I can find the time I will clean it up, if not, hopefully it can be addressed in a subsequent PR!

packages/search-ui/src/DebounceManager.js
```diff
@@ -1,30 +1,58 @@
...
+    const key = functionName + wait.toString();
     let debounced = this.debounceCache[key];
     if (!debounced) {
       this.debounceCache[key] = debounceFn(fn, { wait });
       debounced = this.debounceCache[key];
     }
     debounced(...parameters);
   }
+
+  cancelByName(functionName) {
```
constancecchen: I'd love to see this method get a docblock as well, however short/small!

packages/react-search-ui-views/src/SearchBox.js
JasonStoltz: Sometimes i no tyep good

packages/search-ui/src/__tests__/DebounceManager.test.js
```diff
@@ -0,0 +1,111 @@
+import DebounceManager from "../DebounceManager";
+
+describe("#runWithDebounce", () => {
+  it("will run the provided function after the provided number of milliseconds", () => {
+    const manager = new DebounceManager();
+    let debounced = jest.fn();
```
constancecchen: I don't think we're overriding `debounced` after at any point, are we? Can this be `const debounced` instead of `let`?

packages/search-ui/src/__tests__/DebounceManager.test.js
```diff
@@ -0,0 +1,111 @@
...
+
+    manager.runWithDebounce(1000, "debounced", () => {
```
constancecchen: Correct me if I'm wrong, but can we just do
`manager.runWithDebounce(1000, "debounced", debounced);`
instead of using an anonymous function?

packages/search-ui/src/__tests__/DebounceManager.test.js
constancecchen: Same question as below for all other instances of `let debounced`

packages/search-ui/src/__tests__/DebounceManager.test.js
constancecchen: Same question below for all other instances of
```
() => {
      debounced();
    }
```

packages/search-ui/src/__tests__/DebounceManager.test.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  manager.runWithDebounce(999, "much different", () => {
+    debounced();
+  });
+
+  manager.cancelByName("debounced");
+  jest.advanceTimersByTime(1001);
+  expect(debounced.mock.calls.length).toBe(2);
+});
```
constancecchen: Do we need to write a test for `DebounceManager.debounce` or nah since it's really just calling a 3rd party dep?

packages/search-ui/src/__tests__/DebounceManager.test.js
```diff
@@ -0,0 +1,111 @@
...
+    debounced();
+  });
+
+  manager.runWithDebounce(999, "much different", () => {
+    debounced();
+  });
+
+  manager.cancelByName("debounced");
+  jest.advanceTimersByTime(1001);
+  expect(debounced.mock.calls.length).toBe(2);
```
constancecchen: This test is mixing a couple of different concepts from above (different times+names being distinct), which is great for integration testing, but I'd prefer a smaller one before this for unit testing/dev understanding where the expected calls is 0. Something like:

```js
describe("#cancelByName", () => {
  it("cancels debouncing", () => {
    const manager = new DebounceManager();
    let debounced = jest.fn();

     manager.runWithDebounce(1000, "debounced", () => {
      debounced();
    });

    manager.cancelByName("debounced");
    jest.advanceTimersByTime(1001);
    expect(debounced.mock.calls.length).toBe(0);
  });

  it("cancels functions with different times, but not different names", () => {
    const manager = new DebounceManager();
    let debounced = jest.fn();

     manager.runWithDebounce(1000, "debounced", () => {
      debounced();
    });

     manager.runWithDebounce(500, "debounced", () => {
      debounced();
    });

     manager.runWithDebounce(1000, "different", () => {
      debounced();
    });

     manager.runWithDebounce(999, "much different", () => {
      debounced();
    });

     manager.cancelByName("debounced");
    jest.advanceTimersByTime(1001);
    expect(debounced.mock.calls.length).toBe(2);
  });
});
```

Let me know what you think!

packages/search-ui/src/SearchDriver.js
```diff
@@ -219,85 +222,147 @@ export default class SearchDriver {
...
+
+    // We bail on making state updates if "isLoading" is true, which implies
+    // that there is an active API request that has been made and is pending.
+    //
+    // The net effect of this is that the user interaction effectively becomes
+    // "locked" while API requests are in flight, to avoid any odd race
+    // conditions.
+    //
+    // NOTE: This may be unnecessary now that we have debounce
+    // "_makeSearchRequest".
```
constancecchen: Do we want to consider deleting it in that case? Or leave it in just for now / just in case?

packages/search-ui/src/SearchDriver.js
```diff
@@ -219,85 +222,147 @@ export default class SearchDriver {
       ...searchParameters
     };
 
+    // We want to make sure if this method is running, that any calls to this
+    // method that had been previously deferred are cancelled. We don't want to
+    // possibly have the deferred call execute after this immediate call.
+    this.debounceManager.cancelByName("_updateSearchResults");
```
constancecchen: Is it worth noting in the comments exactly where the debounce for this is started? I had to search through the PR files to realize that it was initiated in setSearchTerm here:

https://github.com/elastic/search-ui/blob/126a3e6503e474a0d3e002cdbaf21249eee70ffc/packages/search-ui/src/actions/setSearchTerm.js#L32-L35

And to be fully honest, I'm still not 100% sure how setSearchTerm is inheriting `this._updateSearchResults` from SearchDriver 😖

packages/search-ui/src/SearchDriver.js
```diff
@@ -219,85 +222,147 @@ export default class SearchDriver {
...
-        if (this.requestSequencer.isOldRequest(requestId)) return;
-        this.requestSequencer.completed(requestId);
+    this._makeSearchRequest({
+      current,
+      filters,
+      resultsPerPage,
+      searchTerm,
+      skipPushToUrl,
+      sortDirection,
+      sortField
```
constancecchen: For keys that are already in state due to `this.setState` above - do we want to consider just pulling those from `this.state` instead of passing `current`, `filters`, `resultsPerPage`, `searchTerm`, `sortDirection`, and `sortField` twice?

packages/search-ui/src/SearchDriver.js
```diff
@@ -219,85 +222,147 @@ export default class SearchDriver {
...
+   *   addFilter("states", "Pennsylvania");
+   *
+   * We don't want to make 3 separate API calls like that in quick succession,
+   * so we debounce the API calls.
+   *
+   * Application state updates are performed in _updateSearchResults, but we
+   * wait to make the actual API calls until all actions have been called.
+   */
+  _makeSearchRequest = DebounceManager.debounce(
+    0,
```
constancecchen: Are we concerned at all about this `0` wait time? Do we want to give it any breathing room, e.g. 3-10, or maybe make it customizable/passable by the developer?

If I remember correctly, 0 just kicks it to the end of the event loop/queue - so if that's all we need, no worries 👍

packages/react-search-ui/src/containers/SearchBox.js
```diff
@@ -71,22 +72,27 @@ export class SearchBoxContainer extends Component {
   };
 
   completeSuggestion = searchTerm => {
-    const { setSearchTerm } = this.props;
-    setSearchTerm(searchTerm);
+    const { clearFilters, setSearchTerm } = this.props;
+    setSearchTerm(searchTerm, {
+      ...(clearFilters === false && { clearFilters })
```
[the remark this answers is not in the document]
JasonStoltz: I was trying to avoid this being passed through:
```
{
  clearFilters: undefined
}
```

I think the only real reason I did that was because it made a bunch of tests fail unintentionally. Probably because they were poorly written tests. I will change this.

packages/search-ui/src/__tests__/SearchDriver.test.js
```diff
@@ -437,3 +441,69 @@ describe("_updateSearchResults", () => {
...
+      },
+      {
+        field: "field3",
+        values: ["value3"],
+        type: "all"
+      }
+    ]);
+  });
+
+  describe("and setSearchTerm is called with a debounce", () => {
```
constancecchen: Nit: is the "and" at the front of this context necessary? I think you can remove it for just "setSearchTerm is called with a debounce" and it still makes sense.

packages/search-ui/src/__tests__/SearchDriver.test.js
```diff
@@ -437,3 +441,69 @@ describe("_updateSearchResults", () => {
...
+          field: "field3",
+          values: ["value3"],
+          type: "all"
+        }
+      ]);
+
+      jest.runAllTimers();
+      // If this case were not working correctly, there would have been a second call here
+      // which would have cleared out the existing filters, since that is the behavior of a "setSearchTerm"
+      // action that is debounced by 1000 milliseconds.
```
constancecchen: This comment was super helpful for me with conceptually understanding what's going on, so really appreciate it! 😅

packages/search-ui/src/test/helpers.js
```diff
@@ -92,12 +94,6 @@ export function doesStateHaveResponseData(response) {
   );
 }
 
-export function waitABit(length) {
```
constancecchen: This feels great to be able to clean out! 🎉

packages/react-search-ui/src/containers/SearchBox.js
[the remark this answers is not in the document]
constancecchen: Ahh interesting. That seems like a scenario which `defaultProps` should be able to handle - or a `clearFilters = true` fallback for non React 🤔

packages/search-ui/src/DebounceManager.js
JasonStoltz: @constancecchen I changed it to an explicit functionName, so that it could be easily cancelled from elsewhere.

Like if I'm running "_updateSearchResults" explicitly, I want to make sure any pending calls to that function are cancelled... if that make sense.

JasonStoltz: By calling ` this.debounceManager.cancelByName("_updateSearchResults");`

packages/search-ui/src/DebounceManager.js
JasonStoltz: Not intentional, I will clean that up.

packages/react-search-ui/src/containers/SearchBox.js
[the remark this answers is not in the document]
JasonStoltz: For sure. I was thinking I'd do the later.

packages/search-ui/src/DebounceManager.js
JasonStoltz: Ah, yes. Good call, thanks for keeping me on my toes!

packages/search-ui/src/DebounceManager.js
constancecchen: ooh yeah, I'm actually not sure how `this.foo` would translate to `.toString`, interesting! 👍 Either way this change seems to work well!

packages/search-ui/src/DebounceManager.js
JasonStoltz: That's a great point! The functions are cached by function name and time. So we could have:

```
{
  someFunction1000: {},
  someFunction500: {},
  someFunction100: {},
}
```

I could simply change it to be:
```
{
  someFunction|1000: {},
  someFunction|500: {},
  someFunction|100: {},
}
```

And do `startsWith(`${functionName}|`).

packages/search-ui/src/DebounceManager.js
JasonStoltz: I wanted to make it a static method, so I didn't have to instantiate a new instance to use it.

I.e., so I could do this:

```
DebounceManager.debounce(...
```

Instead of:
```
const manager = new DebounceManager();
manager.debounce()
```

packages/search-ui/src/__tests__/DebounceManager.test.js
JasonStoltz: Yes, good catch.

packages/search-ui/src/__tests__/DebounceManager.test.js
JasonStoltz: Yes, good catch.

packages/search-ui/src/__tests__/DebounceManager.test.js
JasonStoltz: I didn't just because it was a simple pass-through. But I could for completeness.

packages/search-ui/src/__tests__/DebounceManager.test.js
JasonStoltz: That is a good point, I will do that.

packages/search-ui/src/SearchDriver.js
JasonStoltz: I don't know exactly what to do with that 🤔

packages/search-ui/src/SearchDriver.js
JasonStoltz: Yeah, that's a really good point.

And I'll tell you, but you're not going to like it... one sec.

JasonStoltz: They're mixed in here:

https://github.com/elastic/search-ui/blob/master/packages/search-ui/src/SearchDriver.js#L98-L107

packages/search-ui/src/SearchDriver.js
JasonStoltz: That was my intent, just kicking it to the end of the loop so that you can call multiple actions in a row. I don't actually want to delay processing these by any length of time. Does that make sense?

packages/search-ui/src/DebounceManager.js
constancecchen: Ahhh okay. That gives me a tiny amount of heebly jeeblies but I don't hate it enough to want to change it 👻

packages/search-ui/src/DebounceManager.js
constancecchen: Ahh gotcha! My bad, I hadn't realized we were combining the time with the name for the cache key (although that makes a lot of sense ha).

I think `|` should work really well as a separator since JS functions [can't include that character](https://stackoverflow.com/questions/1661197/what-characters-are-valid-for-javascript-variable-names) in a valid function name! 👍

packages/search-ui/src/__tests__/DebounceManager.test.js
constancecchen: _rubbing hands in anticipation of all the extra lines saved_ :trollface:

packages/search-ui/src/__tests__/DebounceManager.test.js
constancecchen: Depends on how much you care about line/function % coverage. I'm an overachieving completionist when it comes to that stuff but like you said, it's pretty straightforward in this case, so I'm fine with skipping for speed 👍

packages/search-ui/src/SearchDriver.js
constancecchen: Sorry, as in, you're not sure what my comment means? I was mostly responding to the `This may be unnecessary now` sentence, and wondering if we should then delete the line below (the `isLoading` check).

packages/search-ui/src/SearchDriver.js
constancecchen: Oh yeah, that's right, I totally forgot about that! It's been a while haha.

All good, but I think maybe just a short comment above with something like `this is initiated in the setSearchTerm action` would be helpful for developer navigation.

packages/search-ui/src/SearchDriver.js
constancecchen: Gotcha. The only even possible edge case scenario I can think of is a Starcraft player with very high APM clicking 2-3 filters in a row within 5ms of each other, and a developer maybe wanting to batch for that, but that's such a ridiculous edge case you should just ignore me haha.

packages/search-ui/src/SearchDriver.js
JasonStoltz: That is what I'm unsure of, whether or not to delete that line.

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
```diff
@@ -100,20 +126,19 @@ describe("#setSearchTerm", () => {
     expect(subject("term", { refresh: false }).state.wasSearched).toBe(false);
   });
 
-  it("Will not debounce requests if there is no debounce specified", async () => {
+  it("Will not debounce requests if there is no debounce specified", () => {
     const { driver, mockApiConnector } = setupDriver();
     driver.setSearchTerm("term", { refresh: true });
-    driver.setSearchTerm("term", { refresh: true });
-    driver.setSearchTerm("term", { refresh: true });
```
constancecchen: Don't we still want to keep these 2 other calls to make sure that it's not debouncing (length = 3)? Or am I missing something with this test?

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
```diff
@@ -100,20 +126,19 @@ describe("#setSearchTerm", () => {
...
+    jest.runAllTimers();
+    expect(getSearchCalls(mockApiConnector)).toHaveLength(1);
   });
 
-  it("Will debounce requests", async () => {
+  it("Will debounce requests", () => {
     const { driver, mockApiConnector } = setupDriver();
-    driver.setSearchTerm("term", { refresh: true, debounce: 10 });
-    driver.setSearchTerm("term", { refresh: true, debounce: 10 });
-    driver.setSearchTerm("term", { refresh: true, debounce: 10 });
```
constancecchen: Similar-ish question as above - shouldn't we add/keep at least one more `driver.setSearchTerm()` call to ensure debouncing is working properly?

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
```diff
@@ -201,7 +226,8 @@ describe("#setSearchTerm", () => {
         ]
       });
     });
-    it("Will not debounce requests if there is no debounce specified", async () => {
+
```
constancecchen: Awesome catch!

packages/search-ui/src/actions/setSearchTerm.js
```diff
@@ -30,11 +31,16 @@ export default function setSearchTerm(
...
       this._updateSearchResults,
       {
         current: 1,
-        filters: []
+        ...(clearFilters && { filters: [] })
       },
+      // setSearchTerm is meant to be called in quick succession, on every
+      // key stroke in a search box. For this reason, we need to ignore the
+      // "isLoading" check in order to not lock up the UI, as multiple requests
+      // are made in quick succession.
```
constancecchen: Definitely appreciate this new comment block!

packages/search-ui/src/DebounceManager.js
JasonStoltz: Lol, static methods are a common thing though right?

packages/search-ui/src/actions/setSearchTerm.js
```diff
@@ -30,11 +31,16 @@ export default function setSearchTerm(
   if (refresh) {
     this.debounceManager.runWithDebounce(
       debounce,
+      "_updateSearchResults",
       this._updateSearchResults,
       {
         current: 1,
-        filters: []
+        ...(clearFilters && { filters: [] })
```
constancecchen: 🎉 Nice and clean. Phew okay... I think that's it! I've looked through all the files at this point. I think most of my comments were primarily side linting / comment / dev clarity type things. The main functionality is amazing, seriously huge kudos for your brainwork there!!

packages/search-ui/src/SearchDriver.js
constancecchen: Oh hahaha. I mean we could comment it out with a TODO and see what breaks in a month, and if nothing does delete it for reals? :trollface:

packages/search-ui/src/__tests__/SearchDriver.test.js
JasonStoltz: It's easy to get your PRs to go green by commenting out test lines.

packages/search-ui/src/__tests__/DebounceManager.test.js
JasonStoltz: Yeah, this is a much better way to write tests, thank you for pointing that out.

packages/search-ui/src/SearchDriver.js
JasonStoltz: It probably warrants some thorough investigation. Honestly, it's a little entangled in everything. Which is why I'm hesitant to take it on right now and also hesitant to remove it.

constancecchen: Yep, I'm good with that - my comment was mostly tongue in cheek haha.

packages/search-ui/src/SearchDriver.js
JasonStoltz: Yes, I think that would be more clear, and also let _makeSearchRequest callable from other areas of code directly.

packages/search-ui/src/SearchDriver.js
JasonStoltz: @constancecchen I did some analysis of this code and ended up removing this check.

As far as I understand, we had the check in there so that no two concurrent requests were ever being made. This can cause complexity because...
1) If a later request completes before a former, the WRONG search results will be shown on the page
2) If there are 2 requests in flight and one finishes (but not the other), the "isLoading" flag will incorrectly be set to false.

These concerns have actually been addressed by the introduction of the `RequestSequencer` inside of this class.

After removing it, I've done some testing and everything looks good. Scarily, no tests broke when I removed it.

This greatly simplifies the code, I'm glad you pointed it out.

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
JasonStoltz: Fixed in https://github.com/elastic/search-ui/pull/391/commits/159f8c1e6b5afcda13b6ee29acc8fbc946c16d40 with an explanation.

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
JasonStoltz: Fixed in https://github.com/elastic/search-ui/pull/391/commits/159f8c1e6b5afcda13b6ee29acc8fbc946c16d40 with an explanation.

packages/search-ui/src/SearchDriver.js
constancecchen: Awesome!

Quick question - can we remove all instances of isLoading from our state now that we don't seem to be using it to return early?

https://github.com/elastic/search-ui/blob/b0c81b8f5fb9144d10c9801b2e95c637f8cdfce7/packages/search-ui/src/SearchDriver.js#L44

https://github.com/elastic/search-ui/blob/b0c81b8f5fb9144d10c9801b2e95c637f8cdfce7/packages/search-ui/src/SearchDriver.js#L280

https://github.com/elastic/search-ui/blob/b0c81b8f5fb9144d10c9801b2e95c637f8cdfce7/packages/search-ui/src/SearchDriver.js#L311

Unless I'm missing something and it's being passed implicitly down to a child that's using it?...

constancecchen: EDIT: Ah okay we do seem to be passing it down

https://github.com/elastic/search-ui/blob/00013212e7a19ca2e086aa147c53a228eff61fab/ADVANCED.md#L131

In which case nevermind me :) Will mark this as resolved

constancecchen: Awesome. Thanks for explaining that - this looks great!

packages/search-ui/src/__tests__/actions/setSearchTerm.test.js
constancecchen: Awesome. Thanks for explaining that - this looks great!

JasonStoltz:
@constancecchen This is ready for a re-review.

After giving it some thought, I think we could make a really nice refactor to the SearchDriver.

The idea I had is to clearly distinguish "side effects" inside of the driver, and pull them out into their own modules.

For example, running actions could potentially result in 4 different "side effects".

- Updating request state (i.e., "_updateSearchResults")
- Making an API request (i.e., "_makeAPIRequest")
- Updating A11y notifications
- Updating the URL state

packages/search-ui/src/SearchDriver.js
JasonStoltz: @constancecchen Yeah, "isLoading" is actually still important for users. It's let's them hook into our loading state logic and show a loading spinner, etc. if they need to.

JasonStoltz:
For another PR though.

Issue #194, opened 2019-04-17
There should be some way to initiate an initial, empty query when the UI loads.


JasonStoltz:
Here is an example:

When navigating to the sandbox example at http://localhost:3000, currently, you will see no results.
When navigating to the sandbox example at http://localhosts:3000?q=park, you WILL see results, because a 'q' parameter is being passed and a query is performed.

There should be a way to tell Search UI to also trigger a search and show results in the first example, where no 'q' parameter is passed.

A proposal might be to have something like the following, using a "shouldSearchOnInitialLoad" boolean flag in configuration.

```
<SearchProvider
      config={{
        apiConnector: connector,
        shouldSearchOnInitialLoad: true
      }}
    >
      ...
    </SearchProvider>
```
As Constance points out in a comment below, the corresponding change would need to be made here: https://github.com/elastic/search-ui/blob/master/packages/search-ui/src/SearchDriver.js#L160.
 I believe the logic *may* be as simple as adding an additional condition to check if `shouldSearchOnInitialLoad` is true when deciding whether or not to perform a search.

JasonStoltz:
Hi @delvedor :wave:. You had mentioned this earlier, just wanted to point the issue out to you if you'd like to follow it.

constancecchen:
Just wanted to write down some quick notes from a chat I had with @JasonStoltz about this issue today!

- The area to look at for implementation is probably in [searchDriver.js:160](https://github.com/elastic/search-ui/blob/master/packages/search-ui/src/SearchDriver.js#L160), which checks for a `?q=foobar` query param before initializing with results.

- We should consider whether a 0 Results vs All Results state should be the default. I proposed that All Results is more intuitive as a default, and Jason recommended that we still add a config flag that that adds allows a 0 Results state for users that don't want results to be auto-loaded (an example use-case of this would be Google Maps, which doesn't show search results initially).

Kacppian:
I'll pick this up

JasonStoltz:
@Kacppian Great, let us know if you have any questions.

Kacppian:
Sorry for taking a while on this. I think the line no. mentioned might not be accurate? I think the line no is `175`. Please correct me if I'm wrong. 

The conditional checks if a `valid searchTerm` is present or there's at least 1 filter. We'll be replacing the check for a `valid searchTerm` with the flag `shouldSearchOnInitialLoad`, right?

yakhinvadim:
[...]

PR #392, opened 2019-09-20
Initial search

andersosthus:
## Description
When `alwaysSearchOnInitialLoad` is set to `true`, the `SearchDriver` will do an initial search, even if `searchTerm` and/or `filters` is empty.

## List of changes
- Added `alwaysSearchOnInitialLoad` to the config options of `SearchDriver` that defaults to false
- Added a check to the constructor of `SearchDriver`. The check is at the same place as the original `should we do an initial search is` (`SearchDriver.js` line 180)
- Added two tests to verify that `alwaysSearchOnInitialLoad` behaves as expected
- Added documentation for the option `alwaysSearchOnInitialLoad` to `ADVANCED.md`

## Associated Github Issues
#194


JasonStoltz:
@andersosthus This is great! TY for the contribution. I should be publishing this in 1.2 later this week. 👏.

Issue #305, opened 2019-07-02
Replace react-select with something else


JasonStoltz:
> react-select was the only somewhat annoying thing to style over with pure CSS, since it uses inline styling. I’m a bit biased in preferring CSS over CSS-in-JS, but it might be worth considering swapping to another accessible library (e.g. react-responsive-select) that makes our method of overriding styling more coherent.

Additionally, the inline styles cause an issue with CSR.

Relates to #93

Jfelix61:
Hey Jason I'll be trying this!

Jfelix61:
@JasonStoltz Was checking react-responsive-select and it is not able to change its className. Should we change its default styling or change our CSS?

constancecchen:
Aw man, that kinda sucks that they don't pass a custom `className`. If we had to, I'd be fine overriding react-responsive-select's default styling/classes in our CSS files.

Just to throw this out there, one other option (if someone else feels really strongly) could be checking out [EUI's Select component](https://elastic.github.io/eui/#/forms/form-controls)? 🤷‍♀

JasonStoltz:
So what are our big requirements here? I think they are that:
- Should not use inline styles
- Should be accessible

I'd be fine with considering the EUI Select as long as it's possible to use it without requiring all of EUI as a dependency.

yakhinvadim:
I agree with the accessability requirement, but why we don't want the select to use inline styles?
@JasonStoltz 
> the inline styles cause an issue with CSR.
What's the issue?

yakhinvadim:
Hey @Jfelix61, I see that you created a PR replacing react-select with react-responsive-select. I think changing the underlying UI library is a big change for Search UI, what was your reasoning behind choosing react-responsive-select among other libraries? I want to be sure that we're making the right decision here.

yakhinvadim:
I think react-responsive-select in the first message was just an example, and we should research other libraries as well. 
Another option would be to try to improve the accessibility of react-select by making a PR in their repo.

Jfelix61:
Hi @yakhinvadim, I researched other libraries with a screen reader enabled and the one that had the best accessibility was react-select (EUI has good accessibility too, but I can't find if it currently has tree shaking). 
Since we are having problems with CSR and one of the best I could find without inline CSS among the others was react-responsive-select, I decided to implement that one, which in my opinion is not as good on accessibility than react-select but better than others. If you or anyone of the team has more suggestions, I will be glad to change it promptly. :grin:

[...]

PR #410, opened 2019-10-10
[WIP] Replaced react-select with react-responsive-select

Jfelix61:
## Description

Replaced react-select with react-responsive-select to improve accessibility and remove inline styles.

## List of changes

- Removed react-select from ResultsPerPage, Sorting, SingleSelectFacet
- Added DownChevron component for caret icon
- Missing tests, and some styles.

## Associated Github Issues
#305

### DIFF
--- a/examples/elasticsearch/package-lock.json
+++ b/examples/elasticsearch/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "elasticsearch-example",
-  "version": "1.0.0",
+  "version": "1.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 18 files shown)


Jfelix61:
The implementation is almost done. What is missing is a test fail which concerns when values is an empty array (Left it commented out for now). react-responsive-select by default, when no selected value is given, picks the first option available. Therefore, what does the team think of modifying the test to match the first option or should I look for a workaround?

yakhinvadim:
We had a discussion of this change in https://github.com/elastic/search-ui/issues/305#issuecomment-541336543 and decided not to change the select library. 

@Jfelix61, thanks again for your initiative! It's great to have you as a contributor 👍

PR #413, opened 2019-10-17
Feature/issues 411

### LINKED ISSUE #411: Support "any" filters at top level
Use case: Searching for "parks" which are in "California" OR are "World Heritage Sites"

This is currently not possible. The following actions:

```
addFilter("states", "California", "any")
addFilter("states", "Alaska", "any")
addFilter("world_heritage_site", "true", "any")
```

Would result in the query:

```
"filters": {
    "all": [
      {
        "any": [
          { "states": "California" },
          { "states": "Alaska" }
        ]
      },
      {
        "any": [
          { "world_heritage_site": "true" }
        ]
      }
    ]
  }
```

Currently, the "any" parameter affects the relation of a filter to other filters on the same field, NOT the relation to other fields, which is that top level "all".

We could provide a second parameter to choose "all" or "any" at the top level:

```
addFilter("states", "California", "any", "any")
addFilter("states", "Alaska", "any",  "any")
addFilter("world_heritage_site", "true", "any")
```

```
"filters": {
    "any": [
      {
        "any": [
          { "states": "California" },
          { "states": "Alaska" }
        ]
      },
      {
        "any": [
          { "world_heritage_site": "true" }
        ]
      }
    ]
  }
```

A real world use case for this:
> Like while checking package delivery availability one person might want to see if the delivery is available at Office Address OR Home Address

There could be other solutions for this, but this is my current proposal.

In the meantime, users can work around this bye programatically changing the outbound API request: https://github.com/elastic/search-ui/blob/master/ADVANCED.md#api-config

apellizzn:
Add top level configurable filter type `any` `all` and `none`

Resolve #411

### DIFF
--- a/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
+++ b/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
@@ -193,8 +193,8 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "world_heritage_site",
             values: ["true"],
-            type: "all",
-            innerType: "any"
+            outerType: "all",
+            type: "any"
           }
         ],
         sortDirection: "desc",
@@ -300,8 +300,8 @@ describe("AppSearchAPIConnector", () => {
         filters: [
           {
             field: "title",
-            type: "all",
-            innerType: "any",
+            outerType: "all",
+            type: "any",
             values: ["Acadia", "Grand Canyon"]
           },
           {
@@ -321,8 +321,8 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "date_made",
             values: ["yesterday"],
-            type: "all",
-            innerType: "any"
+            outerType: "all",
+            type: "any"
           }
         ]
       };
@@ -498,8 +498,8 @@ describe("AppSearchAPIConnector", () => {
               {
                 field: "world_heritage_site",
                 values: ["true"],
-                type: "all",
-                innerType: "any"
+                outerType: "all",
+                type: "any"
               }
             ],
             sortDirection: "desc",
--- a/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
+++ b/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
@@ -193,7 +193,8 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "world_heritage_site",
             values: ["true"],
-            type: "all"
+            type: "all",
+            innerType: "any"
           }
         ],
         sortDirection: "desc",
@@ -207,7 +208,7 @@ describe("AppSearchAPIConnector", () => {
         filters: {
           all: [
             {
-              all: [
+              any: [
                 {
                   world_heritage_site: "true"
                 }
@@ -300,6 +301,7 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "title",
             type: "all",
+            innerType: "any",
             values: ["Acadia", "Grand Canyon"]
           },
           {
@@ -319,7 +321,8 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "date_made",
             values: ["yesterday"],
-            type: "all"
+            type: "all",
+            innerType: "any"
           }
         ]
       };
@@ -331,7 +334,7 @@ describe("AppSearchAPIConnector", () => {
         filters: {
           all: [
             {
-              all: [
+              any: [
                 {
                   date_made: "yesterday"
                 }
@@ -495,7 +498,8 @@ describe("AppSearchAPIConnector", () => {
               {
                 field: "world_heritage_site",
                 values: ["true"],
-                type: "all"
+                type: "all",
+                innerType: "any"
               }
             ],
             sortDirection: "desc",
@@ -510,7 +514,7 @@ describe("AppSearchAPIConnector", () => {
           filters: {
             all: [
               {
-                all: [
+                any: [
                   {
                     world_heritage_site: "true"
                   }
--- a/packages/search-ui-app-search-connector/src/__tests__/requestAdapters.test.js
+++ b/packages/search-ui-app-search-connector/src/__tests__/requestAdapters.test.js
@@ -24,14 +24,14 @@ const request = {
     {
       field: "initial",
       values: ["values"],
-      type: "all",
-      innerType: "any"
+      outerType: "all",
+      type: "any"
     },
     {
       field: "initial",
       values: ["more values"],
-      type: "all",
-      innerType: "any"
+      outerType: "all",
+      type: "any"
     },
     {
       field: "test",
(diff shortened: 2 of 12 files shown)


yakhinvadim:
Thank you for your PR, @apellizzn! We'll review it and get back to you at the beginning of the next week.

JasonStoltz:
@apellizzn Yep, I believe so!

apellizzn:
@JasonStoltz Updated!

JasonStoltz:
@apellizzn Thanks! I'm digesting this little by little.

A few things:

1) https://github.com/apellizzn/search-ui/blob/feature/issues-411/packages/react-search-ui/src/types/Filter.js and https://github.com/apellizzn/search-ui/blob/feature/issues-411/packages/react-search-ui-views/src/types/Filter.js should be updated

2) The `setFilter` action should be updated to match `addFilter`

3) We need to consider what the behavior should be in the SiteSearchAPIConnector before we can merge this. I'm giving that some thought now.

apellizzn:
@JasonStoltz Changes made, let me know if you think we need anything else.

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -207,7 +208,7 @@ describe("AppSearchAPIConnector", () => {
         filters: {
           all: [
             {
-              all: [
+              any: [
```
JasonStoltz: I'm worried that the meaning of `type` has changed in our `Filter` model.

That parameter previously affected the inner filter, but now there is an `innerFilter` which affects that. This is apparent by looking at this test.

I think you are using the term "innerType" to refer to nested filters:

```
filters: {
  all: [{ // type
    any: {} // innerType
  }]
}
```

Because `type` is already used for nested filters, I think we should update the semantics to this:

```
filters: {
  all: [{ // outertype
    any: {} // type
  }]
}
```

This way we can safely add a `outerType` property without a breaking change, and leave the meaning of `type` in place.

If we do go that route, this test would also need tests to test the `outerType`, whether it's here or in `requestAdapter.test.js`

packages/search-ui-app-search-connector/src/__tests__/requestAdapters.test.js
```diff
@@ -71,13 +76,13 @@ const adaptedRequest = {
   filters: {
     all: [
       {
-        all: [{ initial: "values" }]
+        any: [{ initial: "values" }]
```
JasonStoltz: My previous comment applies to everything here as well.

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -193,7 +193,8 @@ describe("AppSearchAPIConnector", () => {
           {
             field: "world_heritage_site",
             values: ["true"],
-            type: "all"
+            outerType: "all",
```
JasonStoltz: By changing this test case to have `outerType: "all"`, it's hard to tell if anything has broken or not. Previously, omitting `outerType` would default the outer type to "all". Does it still? I can't tell in this test. I think I would actually just remove it from this test and just test the `outerType` behavior in `requestAdapters.test.js`.

packages/search-ui-app-search-connector/src/__tests__/requestAdapters.test.js
```diff
@@ -24,12 +24,14 @@ const request = {
     {
       field: "initial",
       values: ["values"],
-      type: "all"
+      outerType: "all",
```
JasonStoltz: Same thing here, don't update this test, create new cases (Or even just new objects within this test case) to test `outerType`.

packages/search-ui-app-search-connector/src/requestAdapters.js
```diff
@@ -16,16 +16,21 @@ function rollup(f) {
...
 }
 
 function adaptFilters(filters) {
   if (!filters || filters.length === 0) return {};
-  const all = filters.map(rollup);
-  return {
-    all
-  };
+
+  return ["all", "any", "none"].reduce((acc, topType) => {
```
JasonStoltz: We should have tests that cover this logic. For instance, I didn't see anything that tests the "none" case, or a case for values not inside of this list.

packages/search-ui-app-search-connector/src/requestAdapters.js
```diff
@@ -16,16 +16,21 @@ function rollup(f) {
   }));
 
   return {
-    [f.type || "any"]: values
+    [f.type || "all"]: values
```
JasonStoltz: I couldn't reason about why this changed, but I think after you update the tests it might make more sense.

packages/search-ui/src/__tests__/actions/addFilter.test.js
```diff
@@ -64,40 +66,57 @@ describe("#addFilter", () => {
...
-        initialFilters: [{ field: "initial", values: ["value"], type: "all" }]
+      subject("test", "value", "any", undefined, {
+        initialFilters: [
+          { field: "initial", values: ["value"], outerType: "all", type: "any" }
+        ]
       }).filters
     ).toEqual([
-      { field: "initial", values: ["value"], type: "all" },
-      { field: "test", values: ["value"], type: "all" }
+      { field: "initial", values: ["value"], outerType: "all", type: "any" },
```
JasonStoltz: Is there still a test that tests when outerType is not used? I think it may be better to keep this test unchanged and add a new test which includes outerType

packages/search-ui/src/__tests__/actions/addFilter.test.js
```diff
@@ -64,40 +66,57 @@ describe("#addFilter", () => {
...
+      { field: "test", values: ["value"], outerType: "all", type: "any" }
     ]);
   });
 
   it("Adds an additional filter", () => {
     expect(
-      subject("test", "value2", undefined, {
+      subject("test", "value2", "any", undefined, {
```
JasonStoltz: Same here, I'm not sure we should be updating these existing tests.

packages/search-ui/src/__tests__/actions/addFilter.test.js
```diff
@@ -64,40 +66,57 @@ describe("#addFilter", () => {
...
+        outerType: "all",
+        type: "any"
+      }
     ]);
   });
 
   it("Won't add a duplicate filter", () => {
     expect(
-      subject("test", "value", undefined, {
+      subject("test", "value", "any", undefined, {
```
JasonStoltz: Same here, and for the rest of this file.

packages/search-ui/src/__tests__/actions/setFilter.test.js
```diff
@@ -64,54 +66,63 @@ describe("#setFilter", () => {
...
-      { field: "test", values: ["value2"], type: "all" }
+      { field: "initial", values: ["value"], type: "all", outerType: "all" },
+      { field: "test", values: ["value2"], type: "all", outerType: "all" }
     ]);
   });
 
   it("Adds an 'any' type filter", () => {
-    expect(subject("test", "value", "any").filters).toEqual([
-      { field: "test", values: ["value"], type: "any" }
+    expect(subject("test", "value", "any", "any").filters).toEqual([
```
JasonStoltz: I have the same comment for this entire file, it'd be better to add new tests for outerType, I think.

packages/search-ui/src/actions/setFilter.js
```diff
@@ -6,14 +6,23 @@
  * @param name String field name to filter on
  * @param value String field value to filter on
  * @param type String (Optional) type of filter to apply
+ * @param outerType String (Optional) type of inner filter to apply
```
JasonStoltz:
```suggestion
 * @param outerType String (Optional) type of outer filter to apply
```

JasonStoltz:
This PR has gone stale, closing for now.

PR #415, opened 2019-10-23
Design Updates (possibly v2 related)

johnbarrierwilson:
## Description

This is a work-in-progress PR for some various design updates to the react views. To provide proper context, here is a screenshot of the UI design before progress began.

| Screenshot |
| --- |
| PixelSnap 2019-10-23 at 15 18 21@2x |

Below is a thread of design changes, decisions and discussions. Since these changes might end up being broad and possibly breaking, it might be best to make sure this bit of work is aimed for v2.0. Post/comment below if there is anything that you'd like to see updated!

## List of changes

_This list will be updated based on posts below._

Changes include:
- [x] Limit width of search bar
- [x] Modify colors to remove unnecessary borders


johnbarrierwilson:
Here's the first round of updates that make two large changes:
- Limit the width of the search bar
- Modify colors to remove unnecessary borders

PixelSnap 2019-10-23 at 15 22 47@2x

JasonStoltz:
@johnbarrierwilson Hey John, you should PR this into master, not stable, fyi.

JasonStoltz:
@yakhinvadim What is your opinion on this, do we consider this a breaking change?

Personally, I do.

The reason is this, for example. I'd like to keep the Search UI demo page up to date with the latest version of Search UI: https://swiftype.com/search-ui.

I would expect an update from 1.2 to 1.3 to be seamless.

However, since we override the default styles in that demo, I can almost guarantee that any stylesheet changes will mean I'll have to manually adjust my styles to compensate.

For example:

In the original stylesheet. These are for the text input button:
```
padding: 6px;
margin: 0;
```

Changed to:
```
padding: 8px 10px;
```

Even though I'm overriding styles, I didn't override the padding style on the text input, so after installing 1.3, my text input will now be larger. It may end up throwing off the entire layout, right? I mean depending on what styles I modified, I could have been relying on that button being an exact size, and now that it's larger and potentially wrapping.

For these to NOT be breaking, I think they'd need all be off of a new class that you could opt into. Like `.v2` or something.

yakhinvadim:
Good example, Jason. This is definitely a breaking change.

johnbarrierwilson:
I'm totally on board with making this stuff `v2`. I'll continue to work on this since it's about 50% thought through at the moment.

JasonStoltz:
Closing this for now. Feel free to reopen at some point.

Issue #436, opened 2019-11-08
Site Search spelling suggestions are un available in the response


JasonStoltz:
Original question:

> Hi there!
> My use case needs spelling suggestions from a site search engine. I can alter the query options:

```
connector = new SiteSearchAPIConnector({
    engineKey: xxxx,
    documentType: xxxx    
    beforeSearchCall: (queryOptions, next)=> {
      const withSpelling = {...queryOptions, spelling: 'strict'}
      return next(withSpelling)
    }
})
```

> And I can see the spelling suggestion coming back in the info field of the response but I don't see any way to surface this in search-ui. I'm guessing this isn't supported right now but if I've missed something obvious, please let me know :)

We unfortunately have no work-around for this. I think we could likely easily add a "didYouMean" field to state, though, which the site search connector could populate with this spelling suggestion.

Because it is a new field, it would be a non-breaking change. This feature is not supported by app search but could easily be added in the future if that feature becomes available.

We would also probably want to add search query configuration for this as well. I am imagining a configuration like:

```
config = {
  searchQuery: {
    didYouMean: { // Using an object to leave room for additional config in the future
      enabled: true
    }
  }
}
```

Essk:
I'm wondering if it would be easier to build the request if the config directly mirrors the site-search options: 'strict', 'always', 'retry'?

```
config = {
  searchQuery: {
    didYouMean: { // Using an object to leave room for additional config in the future
      enabled: true
      spelling: ['strict', 'always', 'retry']
    }
  }
}
```

JasonStoltz:
@Essk Yes, something like that makes sense to me.

Maybe in that case it would be better to do something like:

```
config = {
  searchQuery: {
    spelling: {
      type: 'strict' // 'strict', 'always', 'retry'
    }
  }
}
```

davidsauntson:
Can I re-open this issue please? As I would like to finish this off in a fork of @Essk fork.  Thanks!

JasonStoltz:
Yep, thanks!

JasonStoltz:
Closed by #517

PR #439, opened 2019-11-14
Feature/site search spelling

Essk:
## Description
Allows users of the site search connector to configure a `spelling` field to request spelling suggestions from the site search API. Any suggestion from the API is then exposed in the response state.

## List of changes
See commits

## Associated Github Issues
https://github.com/elastic/search-ui/issues/436


ADVANCED.md
```diff
@@ -1151,6 +1152,7 @@ For example, if you add a `search_fields` configuration option, it will control
...
 | `result_fields`                    | Object[String, Object]   | optional  |         | Fields which should be returned in results.<br/><br/>[App Search result_fields API Reference](https://swiftype.com/documentation/app-search/api/search/result-fields)                                                                                                                                                                                                                                                          |
+| `spelling`                    | Object  | optional  |         | **Only supported by Site Search** <br/> [Site Search API reference](https://swiftype.com/documentation/site-search/searching/spelling) <br/> Specify the type of spelling behaviour with a `type` field that should have a value of "strict", "always" or "retry" <br/> Example: ```spelling: { type: "strict" }``` |
```
JasonStoltz: Seeing this spelled out in the docs is making me second guess the approach here. It's a bit awkward to have to specify that it is a Site Search only API. I'm also worried about cementing this as part of the API.

I wonder if we could go back to your original approach:
```
    beforeSearchCall: (queryOptions, next)=> {
      const withSpelling = {...queryOptions, spelling: 'strict'}
      return next(withSpelling)
    }
```

Perhaps it would be interesting to change our connectors to store the original, unparsed response in state.

```
state.rawResponse
```

In that case, the AppSearch and SiteSearch connectors would just need to be updated to populate that state field in the response adapters.

It's kind of a fallback; like if all else fails, you can always access the raw response in state. This would solve your case

Another option would be open up a "responseMeta" field in state, which is available for writing arbitrary data to for Connectors. The SiteSearchConnector, in this case, could write out spelling corrections there.

```
state.responseMeta
```

Essk: Hi, Thanks for the feedback. I agree that this current approach raises some questions, I found myself wondering if I should be adding UI component support, but then realising that would not work with an app search API...

For my purposes at least, having the raw response available would solve the problem so I'd be happy to close this PR and implement that instead

JasonStoltz: I'm going to close this PR for now, I would be happy to see a new PR from you.

PR #442, opened 2019-11-19
fix: Replace debounce-fn in DebounceManager to facilitate IE11 compat…

AndrewTziAnChan:
Even after adding polyfills in support of IE11, search-ui is still unable to render properly due to an issue in using Object.defineProperty() when mimic-fn a function as part of the use of debounce-fn.

On our own fork of search-ui, we've replaced the use of the debounce-fn module with our own debounce function. After discussion with my team, we wanted to submit a PR with our debounce function so that we wouldn't have to maintain our own fork specifically for IE11 compatibility.

### DIFF
--- a/packages/search-ui/src/DebounceManager.js
+++ b/packages/search-ui/src/DebounceManager.js
@@ -6,7 +6,7 @@
  */
 function debounce(func, wait) {
   let timeout;
-  return function() {
+  const debouncedFn = function() {
     const args = arguments;
     const later = () => {
       func.apply(null, args);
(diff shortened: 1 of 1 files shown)


JasonStoltz:
@AndrewTziAnChan Thank you for the PR. I will review this in the next day or two.

andrew-chernous:
Hi @JasonStoltz  Can you please review PR I really need this fix.

JasonStoltz:
@AndrewTziAnChan Sorry about that. I've been meaning to push a 1.3, so let me merge this up today and push a new release.

katydecorah:
Thank you! This is a big help to our team, too!

PR #443, opened 2019-11-19
fix: Use browser history replace instead of push on first load

### LINKED ISSUE #434: Search-UI smart url forces extra history push
When going to a page with Swiftype, there is an extra browser history push on component load.

For example, on the [CodeSandbox ](https://codesandbox.io/s/national-parks-example-kdyms?fontsize=14)demo on the Readme for this project, the url navigates from

**https://kdyms.csb.app/**

to

**https://kdyms.csb.app/?size=n_20_n**

A browser back navigation action can be triggered and the url will change from **https://kdyms.csb.app/?size=n_20_n** to **https://kdyms.csb.app/** which does not seem like ideal behavior.

If I navigated from **google.com** to **https://kdyms.csb.app/**, I'd expect hitting the back button on the browser from **https://kdyms.csb.app/?size=n_20_n** to navigate back to **google.com**, not to **https://kdyms.csb.app/**.

AndrewTziAnChan:
The idea for this change is to not push browser history on the very first load. Our application uses react-router, which will capture the changes to the url. This will require the user to make 2 back navigations in order to get back to the page they entered the search page rather than just one.

Additionally, if there's some filter state captured in the url, the first of the backs will clear our the initial filter parameters, potentially leading to some odd search results. (Example, we filter out results by environment dev/stage/prod, this defect causes the environment filter to be cleared, showing results from all environments).

Fixes #434

### DIFF
--- a/packages/search-ui/src/URLManager.js
+++ b/packages/search-ui/src/URLManager.js
@@ -117,8 +117,12 @@ export default class URLManager {
 
   pushStateToURL(state) {
     const searchString = stateToQueryString(state);
+    const navigationFunction =
+      this.lastPushSearchString.length === 0
+        ? this.history.replace
+        : this.history.push;
     this.lastPushSearchString = searchString;
-    this.history.push({
+    navigationFunction({
       search: `?${searchString}`
     });
   }
--- a/packages/search-ui/src/URLManager.js
+++ b/packages/search-ui/src/URLManager.js
@@ -122,9 +122,10 @@ export default class URLManager {
         ? this.history.replace
         : this.history.push;
     this.lastPushSearchString = searchString;
-    navigationFunction({
-      search: `?${searchString}`
-    });
+    navigationFunction &&
+      navigationFunction({
+        search: `?${searchString}`
+      });
   }
 
   onURLStateChange(callback) {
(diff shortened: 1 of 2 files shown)


JasonStoltz:
@AndrewTziAnChan Thank you for the PR. I will review this in the next day or two.

mannieg:
@JasonStoltz Hi do you have an idea when you will be able to review this? We are experiencing the same issue that this merge request will resolve.

Thanks again.

JasonStoltz:
@mannieg I'll do it right now and try to get something merged and deployed in a 1.3 release.

JasonStoltz:
Hi, I took a slightly different approach for this fix. Thank you for the contribution though, it definitely got me thinking in the right way about this. Please take a look at this PR and let me know if it makes sense to you: https://github.com/elastic/search-ui/pull/449

JasonStoltz:
@AndrewTziAnChan

AndrewTziAnChan:
@JasonStoltz I'm currently on vacation now and can't completely verify that the changes in your PR would address our issue, but after reviewing the changes, I think they make sense and they look clean. Once I get back in office, I'll be sure to integrate the latest into our project.

PR #449, opened 2019-12-17
fix: Search-UI smart url forces extra history push

JasonStoltz:
Currently, when showing search results on page load, an extra history
event is added to history, causing a few issues, including the
inconvenience on users of having to press the back button twice to get
back to where they were originally.

The fix is to simply use a browser history "replace" instead of "push"
on the any queries happening on page load.

This PR is based on the following PR:
https://github.com/elastic/search-ui/pull/443

Fixes Fixes https://github.com/elastic/search-ui/issues/434

latest

### DIFF
--- a/packages/search-ui/src/SearchDriver.js
+++ b/packages/search-ui/src/SearchDriver.js
@@ -185,7 +185,7 @@ export default class SearchDriver {
       searchParameters.filters.length > 0 ||
       this.alwaysSearchOnInitialLoad
     ) {
-      this._updateSearchResults(searchParameters);
+      this._updateSearchResults(searchParameters, { replaceUrl: true });
     }
   }
 
@@ -217,7 +217,10 @@ export default class SearchDriver {
   /**
    * This method is used to update state and trigger a new search.
    */
-  _updateSearchResults = (searchParameters, { skipPushToUrl = false } = {}) => {
+  _updateSearchResults = (
+    searchParameters,
+    { skipPushToUrl = false, replaceUrl = false } = {}
+  ) => {
     const {
       current,
       filters,
@@ -249,7 +252,8 @@ export default class SearchDriver {
     });
 
     this._makeSearchRequest({
-      skipPushToUrl
+      skipPushToUrl,
+      replaceUrl
     });
   };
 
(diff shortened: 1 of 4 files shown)


packages/search-ui/src/SearchDriver.js
```diff
@@ -249,7 +252,8 @@ export default class SearchDriver {
     });
 
     this._makeSearchRequest({
-      skipPushToUrl
+      skipPushToUrl,
+      replaceUrl
```
constancecchen: Just to make sure I fully understand this functionality: why add a new `replaceUrl` option instead of just making use of the existing `skipPushToUrl` opt? Is it because we do want the URL bar to update?

packages/search-ui/src/SearchDriver.js
```diff
@@ -274,87 +278,91 @@ export default class SearchDriver {
    * Application state updates are performed in _updateSearchResults, but we
    * wait to make the actual API calls until all actions have been called.
    */
```
constancecchen: Actually if there is on request I could make - it'd be great to add some `@params` docblocks here to more fully document the new option, e.g.

```
* @param {int} debounceInterval - ...etc
* @param {obj} options
* @param {bool} options.skipPushToUrl - Skips updating both the URL bar and browser history
* @param {bool} options.replaceUrl - Updates the URL bar, but does not insert an entry in the browser history
```

Something like that would probably help a lot with a new developer quickly grokking the difference between the two flags :)

packages/search-ui/src/SearchDriver.js
JasonStoltz: We want the URL to update, but we don't want a new history event to be added.

packages/search-ui/src/SearchDriver.js
JasonStoltz: Great points.

JasonStoltz: Updated here: https://github.com/elastic/search-ui/pull/449/commits/f60d07f3e241b922903339ad05610fc7e35ff305

JasonStoltz:
@constancecchen

>  it'd be that it sorta feels like we're implementing very general fuctionality for what appears to be a specific problem

Valid point. This seemed to be the cleanest solution to me though.

> Is this an option we're making available to dev users? (i.e., URLManager?) If so, does it need to documented?

I don't think there is any change to the public interface that a dev user would need to know about. URLManager is used internally. I put JSDoc on it if that helps.

> Could this could potentially be misused in the future? Seems super unlikely/I'm probably just being paranoid, I'm guessing the end-user scenario we don't want is actual searches/search changes not leaving browser history as expected, correct?

I don't think so 🤔.

PR #457, opened 2020-01-13
Added ResultWithRequiredId type

### LINKED ISSUE #444: propTypes on Result view don't properly validate incoming data
This causes much confusion:

https://github.com/elastic/search-ui/issues/441

For reference, this is when we loosened the schema: https://github.com/elastic/search-ui/commit/c68f78f9f44680f8219c8aa1e2ed5442c56936af

At minimum, on the view we should validate that an "id" field is present

JasonStoltz:
Some views actually required an "id" property on result objects
to function properly, which was not expressed in validations.

I couldn't simply turn on an "id" requirement for Result across
the board because it could break certain views that actually
don't require it currently.

For that reason, I create a separate view, ResultWithRequiredId, which
enforces the "id" requirement.

The views will still error if an "id" is missing, but at least now a proper error message will be shown in the console.

I also added a special "TODO Next Major" to help indicate breaking
changes that should be made in the next major.

React_App

## Associated Github Issues

Fixes #444

### DIFF
--- a/packages/react-search-ui/src/containers/Results.js
+++ b/packages/react-search-ui/src/containers/Results.js
@@ -4,7 +4,7 @@ import { Result, Results } from "@elastic/react-search-ui-views";
 
 import { withSearch } from "..";
 import { Result as ResultContainer } from ".";
-import { Result as ResultType } from "../types";
+import { ResultWithRequiredId as ResultType } from "../types";
 
 function getRaw(result, value) {
   if (!result[value] || !result[value].raw) return;
--- a/packages/react-search-ui/src/types/Result.js
+++ b/packages/react-search-ui/src/types/Result.js
@@ -2,15 +2,10 @@ import PropTypes from "prop-types";
 
 import FieldValueWrapper from "./FieldValueWrapper";
 
-// Typically an object where keys are the field names, and values are field values.
-// Also could be literally any other arbitrary value depending on the particular Search API response.
-// Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
-// is usually better to work with FieldValueWrapper values. We encourage FieldValueWrapper, but we accept
-// anything because we don't want users to have type error warnings unnecessarily.
+// If field is a "FieldValueWrapper", it can be displayed by the UI.
 //
-// An example would be if a user requests "grouping" in an App Search API request. That will come back
-// as "_group: {..}". It *should* be there in the Result so that a developer has it available to work
-// with.
+// If it is any other object ("PropTypes.any"), it will be passed along with
+// the object but will be ignored by the UI.
 export default PropTypes.objectOf(
   PropTypes.oneOfType([PropTypes.any, FieldValueWrapper])
 );
--- a/packages/react-search-ui/src/types/ResultWithRequiredId.js
+++ b/packages/react-search-ui/src/types/ResultWithRequiredId.js
@@ -1,0 +1,18 @@
+import PropTypes from "prop-types";
+
+/*
+  [TODO Next Major]
+
+  This file exists alongside Result.js to enforce that Result objects have an
+  ID in certain views that already require it.
+
+  I wanted to add id directly to Result.js, but that would be a breaking change. This exists
+  to give better validation warnings to those files that already implicitly required it.
+*/
+export default PropTypes.objectOf(
+  PropTypes.shape({
+    id: PropTypes.shape({
+      raw: PropTypes.string
+    }).isRequired
+  }).isRequired
+);
--- a/packages/react-search-ui-views/src/Autocomplete.js
+++ b/packages/react-search-ui-views/src/Autocomplete.js
@@ -1,7 +1,7 @@
 import PropTypes from "prop-types";
 import React from "react";
 
-import { Result } from "./types";
+import { ResultWithRequiredId as Result } from "./types";
 import { Suggestion } from "./types";
 import { appendClassName } from "./view-helpers";
 
--- a/packages/react-search-ui-views/src/types/Result.js
+++ b/packages/react-search-ui-views/src/types/Result.js
@@ -2,15 +2,10 @@ import PropTypes from "prop-types";
 
 import FieldValueWrapper from "./FieldValueWrapper";
 
-// Typically an object where keys are the field names, and values are field values.
-// Also could be literally any other arbitrary value depending on the particular Search API response.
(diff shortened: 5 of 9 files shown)


packages/react-search-ui/src/types/ResultWithRequiredId.js
```diff
@@ -0,0 +1,18 @@
...
+
+/*
+  [TODO Next Major]
+
+  This file exists alongside Result.js to enforce that Result objects have an
+  ID in certain views that already require it.
+
+  I wanted to add id directly to Result.js, but that would be a breaking change. This exists
+  to give better validation warnings to those files that already implicitly required it.
+*/
```
constancecchen: So just to check my understanding:

1. Are we going back to `Result` (with a required ID) the next major/breaking release?

2. What (if anything) is still using `Result` (vs `ResultWithRequiredId`)? (This is just me being curious)

packages/react-search-ui/src/types/Result.js
```diff
@@ -2,15 +2,10 @@ import PropTypes from "prop-types";
...
-// Typically an object where keys are the field names, and values are field values.
-// Also could be literally any other arbitrary value depending on the particular Search API response.
-// Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
-// is usually better to work with FieldValueWrapper values. We encourage FieldValueWrapper, but we accept
-// anything because we don't want users to have type error warnings unnecessarily.
+// If field is a "FieldValueWrapper", it can be displayed by the UI.
 //
-// An example would be if a user requests "grouping" in an App Search API request. That will come back
-// as "_group: {..}". It *should* be there in the Result so that a developer has it available to work
-// with.
```
constancecchen: It's kind of a shame to lose all this context - the example in particular is helpful. Will we be adding it back at any point?

packages/react-search-ui/src/containers/Results.js
```diff
@@ -4,7 +4,7 @@ import { Result, Results } from "@elastic/react-search-ui-views";
 
 import { withSearch } from "..";
 import { Result as ResultContainer } from ".";
-import { Result as ResultType } from "../types";
+import { ResultWithRequiredId as ResultType } from "../types";
```
constancecchen: OK, so just so I'm understanding the shape change:

`FieldValueWrapper` type:
```
{
  raw: 'foo',
  snippet: 'bar'
}
```

the new WithRequiredId type:
```
{
  id: {
    raw: 'foo',
    snippet: 'optional',
  }
}
```

is that right?

JasonStoltz:
@constancecchen For more context.

The change I wanted to make is to validate across the board that every result coming from the server has `.id.raw`. That's really the only hard requirement that any result absolutely needs to have.

**Exhibit A - The desired shape**

As I was typing a response to you I realized I think I have my shape wrong:

```
export default PropTypes.objectOf(
  PropTypes.shape({
    id: PropTypes.shape({
      raw: PropTypes.string
    }).isRequired
  }).isRequired
);
```

Should be:
```
export default PropTypes.shape({
  id: PropTypes.shape({
    raw: PropTypes.string.isRequired
  }).isRequired
})
```

**Exhibit B - The concern about breaking changes**

In the UI, we have the following components that consume results:

/react-search-ui/src/containers/Result.js - Uses result.raw.id
/react-search-ui-views/src/Result.js
/react-search-ui/src/containers/Results.js - Uses result.raw.id
/react-search-ui-views/src/Result.js
/react-search-ui/src/containers/SearchBox.js
/react-search-ui-views/src/Autocomplete.js - Uses result.raw.id

So currently, if you're using any of the components labeled `Uses result.raw.id` and the server is not returning a `result.raw.id`, they are breaking for you. The changes in this PR simply add a more descriptive error as to why they are breaking.

(To note, the only way you'd get yourself in this situation is if you wrote your own backend connector.)

My concern about adding my desired state shape change across the board if you were currently using a component like `/react-search-ui-views/src/Result.js` directly, your code would run with no errors. If we introduced this new validation, this code would start erroring.

It's an edge case though. These things are typically used in pairs. `/react-search-ui/src/containers/Result.js` is paired with `/react-search-ui-views/src/Result.js`, you wouldn't use `/react-search-ui-views/src/Result.js` directly, although you could.

And actually, propTypes are only checked in development mode and wouldn't break production builds.

So honestly, we could probably make this change across the board and be just fine. I don't know though, what do you think? 

**Exhibit C - The idea of FieldValueWrapper**

```
export default PropTypes.objectOf(
  PropTypes.oneOfType([PropTypes.any, FieldValueWrapper])
);
```

Again, as I mentioned above, the only thing that's REALLY required on a result is `.id.raw`.

The definition for Result currently looks like the above. It's kind of silly. It's basically like, "A result can have literally anything as a property, but it could also have a FieldValueWrapper". So the FieldValueWrapper check is completely unnecessary for type checking, it's more for documentation. 

Honestly, I think it just confuses things and makes this code harder to work with. I think I'd like to remove it.

Any thoughts on all of this?

byronhulcher:
> I think I could just change types/Result.js to the following and make everything a whole lot simpler:

This seems like the right approach ultimately.  This is what will be done next major version? If that's the case I think this is fine to merge as-is, even if it feels weird.  I think semantic versioning compatibility is very important and I'd rather have weird bridge code for a few minors if it means we get into the habit of releasing breaking changes to our code only in majors.

JasonStoltz:
@byronhulcher That's fine, maybe I'll just log an issue to do this the "Right way" in 2.0.

packages/react-search-ui/src/types/Result.js
JasonStoltz: I can add it back. I actually read it back to myself and I thought it sounded longwinded and confusing which is why I removed it.

constancecchen:
I'm good with bridge code for now and the actual final `types/Result.js` update in a major release later.

That explanation was **super** helpful, thanks a ton for writing that all out Jason. Sounds like maybe we need to fix the propTypes shape (I didn't actually even notice it was incorrect so glad you caught that!) and then this is good to go!

packages/react-search-ui/src/types/Result.js
constancecchen: Totally up to you, I'm good either way!

JasonStoltz:
Putting this on hold until 7.6 testing is done.

JasonStoltz:
I'm not happy with this PR, I'm going to close it.

Issue #289, opened 2019-06-26
Add Typescript bindings


JasonStoltz:
Someone will want these, and it will also help us document our own API.

For reference: https://github.com/elastic/eui/commit/1c837b7d1e5dd1371c9e5c5a3733a437b034bd09#diff-f8ac36caa03f828bf516460a9156f08c

harsimranb:
I am planning on using this package in a large TypeScript project, and typescript bindings will be very helpful.

alexanderatallah:
It's been a couple months - just wanted to check if this is being worked on! It's a big part of our decision to use it. Right now, we have to type out things like "Result" manually, otherwise it's very easy to make errors (e.g. when a field is null in our elasticsearch index, or when results aren't being serialized correctly).

JasonStoltz:
@alexanderatallah It is not being worked on currently, but I understand the need.

alexanderatallah:
@JasonStoltz I submitted some type definitions for react-search-ui. Do these look ok? https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41605

JasonStoltz:
@alexanderatallah 

That's great! A few questions... I'm new to TypeScript.

- Did you generate these or hand code them? If generated, how did you generate them?
- I started doing some work to add these definitions directly to this project by adding a `.d.ts` file directly to the top of each package. If I do this, what happens to the types defined in the `@types`? Do they become obsolete?

JasonStoltz:
@alexanderatallah I have a Draft PR up for what I have so far: https://github.com/elastic/search-ui/pull/458. I wonder if it would make more sense to add your changes to this?

alexanderatallah:
@JasonStoltz I hand-coded the types. Yeah it’s a little bit better to add the definitions to the main repo, since it’s more likely that they’ll stay in sync. Not sure about this but adding a tslint command call to the tests might be able to tell when something new got exported that isn’t typed.

I’ll open a PR against yours

JasonStoltz:
This is already in progress

radhikanarang24:
Hey
Just wanted to check if we have TypeScript support available with Search UI?

JasonStoltz:
Not yet

subsocialdev:
> This is already in progress

Hey @JasonStoltz, would you mind sharing with us the progress on adding TypeScript support, please?

If it's not ready, what is the best working draft so we could start using it today?

JasonStoltz:
[...]

PR #458, opened 2020-01-15
WIP: Added Typescript Definitions to search-ui-app-search-connector

JasonStoltz:
## Description

@alexanderatallah  and I have I added TypeScript declarations to the `search-ui-app-search-connector` and `react-search-ui` packages.

NOTE: The following is true of the `search-ui-app-search-connector`, but not yet `react-search-ui`.

The declarations are generated from JSDoc in source. This hopefully helps:
- Readability of code from JSDoc comments
- JSDoc comments for non-TS users
- TS declarations for TS users which can be synced with the JSDoc with somewhat minimal effort

The JSDoc is also being used to generate documentation in READMEs, for certain projects.

## List of changes

- Updated PR template to add a Reminder to update TS declarations
- Added instructions for maintaining declarations
- Added instructions for generating documentation from JSDoc

latest

## Testing

`npm install @elastic/search-ui-app-search-connector@canary`

## Associated Github Issues

https://github.com/elastic/search-ui/issues/289

### DIFF
--- a/packages/react-search-ui/react-search-ui.d.ts
+++ b/packages/react-search-ui/react-search-ui.d.ts
@@ -1,0 +1,345 @@
+declare module "@elastic/react-search-ui" {
+
+  import { Component } from 'react';
+
+  type func = (...args: any[]) => any;
+  type renderFunc = (...args: any[]) => JSX.Element;
+
+  /**
+   * FIELDS
+   */
+  export type FieldValue = string | number | boolean | Array<string | number | boolean>;
+
+  export interface FieldValueWrapper {
+    // A raw field value, like 'I am a raw result', or 2, or true. Raw values may
+    // or may not be html escaped, so *always* sanitize a raw value before rendering
+    // it on a page as html.
+    raw?: FieldValue;
+    // A snippet value contains a highlighted value. I.e., 'I <em>am</em> a raw
+    // result'. These are always sanitized and safe to render as html.
+    snippet?: string;
+  }
+
+  /**
+   * FACETS
+   */
+  export type FacetType = "range" | "value";
+
+  export interface FacetValue {
+    // Number of results for this filter
+    count: number;
+    // Filter to apply if selected
+    value: FilterValue;
+    // Whether or not this facet value is selected
+    selected?: boolean;
+  }
+
+  interface FacetT {
+    data: FacetValue[];
+    // Name of the field this facet is associated with
+    field: string;
+    type: FacetType;
+  }
+
+  /**
+   * FILTERS
+   */
+  export interface Filter {
+    field: string;
+    values: FilterValue[];
+    type: FilterType;
+  }
+
+  export type FilterType = "all" | "any" | "none";
+
+  export type FilterValueValue = FieldValue;
+
+  export interface FilterValueRange {
+    // Beginning of the range, like 1
+    from?: FieldValue;
+    // A unique name for this range, used for display
+    name: string;
+    // End of the range, like 100
+    to?: FieldValue;
+  }
+
+  export type FilterValue = FilterValueRange | FilterValueValue;
+
+  /**
+   * RESULTS
+   */
+  // Typically an object where keys are the field names, and values are field values.
+  // Also could be literally any other arbitrary value depending on the particular Search API response.
+  // Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
+  // is usually better to work with FieldValueWrapper values.
+  // We only accept FieldValueWrapper to help typing
+  //
+  // An example would be if a user requests "grouping" in an App Search API request. That will come back
+  // as "_group: {..}". It *should* be there in the Result so that a developer has it available to work
+  // with.
+  interface ResultT {
+    [key: string]: FieldValueWrapper;
+  }
+
+  /**
+   * SORTING
+   */
+
+  export type SortDirection = "asc" | "desc";
+
+  export interface SortOption {
+    // A display name, like "Name"
+    name?: string;
+    // A field name, like "name".
+    value?: string;
+    // asc or desc
+    direction?: SortDirection | "";
+  }
+
+  export interface Suggestion {
+    suggestion?: string;
+    highlight?: string;
+    data?: object;
+  }
+
+  export interface AutocompleteSection {
+    sectionTitle?: string;
+  }
+
+  /**
+   * CONTAINERS
+   */
+  interface ErrorBoundaryProps {
+    // Props
+    children: JSX.Element;
+    className?: string;
+    view?: renderFunc;
+    // State
+    error?: string;
+  }
+
+  export class ErrorBoundary extends Component<ErrorBoundaryProps> {}
+
+  interface FacetProps {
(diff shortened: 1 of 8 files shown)


packages/search-ui-app-search-connector/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "Apache-2.0",
   "main": "lib",
   "module": "es",
+  "types": "search-ui-app-search-connector.d.ts",
```
alexanderatallah: small remark: I haven't seen this before when adding type declarations - usually it's just an `index.d.ts` file in the root directory

alexanderatallah: small remark: I haven't seen this before when adding type declarations - usually it's just an `index.d.ts` file in the root directory

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
```diff
@@ -0,0 +1,65 @@
+declare module "@elastic/search-ui-app-search-connector" {
+  class AppSearchAPIConnector {
+    /**
+     * @typedef {Object.<string, *>} KeyValuePairs
+     */
+    /**
+     * @callback next
+     * @param {KeyValuePairs} updatedQueryOptions The options to send to the API
+     * @returns {Object} The response
```
alexanderatallah: in typescript, there's no need for typing like `{Object}` in the comments, as those are a part of the types themselves

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
```diff
@@ -0,0 +1,65 @@
...
+      beforeSearchCall,
+      beforeAutocompleteResultsCall,
+      beforeAutocompleteSuggestionsCall,
+      endpointBase
+    }: {
+      searchKey: string;
+      engineName: string;
+      hostIdentifier?: string;
+      beforeSearchCall?: (
+        queryOptions: {
```
alexanderatallah: `{ [x: string]: any }` is equivalent to `object`. Usually the latter is preferred

packages/search-ui-app-search-connector/package.json
JasonStoltz: Thanks @alexanderatallah, I was somewhat following our eui project as an example: https://github.com/elastic/eui/blob/master/package.json#L8

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
JasonStoltz: @alexanderatallah That's fair, I think I will make that change. So type `Object` in uppercase, or `object` in lower case?

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
JasonStoltz: @alexanderatallah These comments were copied over when running the `typescript` command to generate the definitions. I could disable that since the comments are included in the source files, but I thought maybe having them here as well would add some helpful context.

JasonStoltz: And they are required in the source file, because a lot of the parameter typing in this is actually generated from the JSDoc in source.

packages/search-ui-app-search-connector/package.json
alexanderatallah: Sounds good!

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
alexanderatallah: Ah ok!

packages/search-ui-app-search-connector/search-ui-app-search-connector.d.ts
alexanderatallah: @JasonStoltz `object` in lower case is better. According to https://github.com/DefinitelyTyped/DefinitelyTyped:
> Using the types Function and Object is almost never a good idea. In 99% of cases it's possible to specify a more specific type. Examples are (x: number) => number for functions and { x: number, y: number } for objects. If there is no certainty at all about the type, any is the right choice, not Object. If the only known fact about the type is that it's some object, use the type object, not Object or { [key: string]: any }.

packages/search-ui-app-search-connector/package.json
alexanderatallah: Sounds good!

JasonStoltz:
@constancecchen Or anyone else that was planning to review this, please hold off, I'm going to do some additional cleanup after merging another PR into this one.

packages/react-search-ui/react-search-ui.d.ts
```diff
@@ -0,0 +1,345 @@
...
+  export interface FacetValue {
+    // Number of results for this filter
+    count: number;
+    // Filter to apply if selected
+    value: FilterValue;
+    // Whether or not this facet value is selected
+    selected?: boolean;
+  }
+
+  interface FacetT {
```
alexanderatallah: In my own code, I found that I actually needed this type for something - think we should export it like the rest? (generally, the rule is to export everything, I just didn't originally because the source code's class is named `Facet`)

packages/react-search-ui/react-search-ui.d.ts
```diff
@@ -0,0 +1,345 @@
...
+  // Typically an object where keys are the field names, and values are field values.
+  // Also could be literally any other arbitrary value depending on the particular Search API response.
+  // Default views in Search UI know what to do with FieldValueWrapper values, but not arbitrary values, so it
+  // is usually better to work with FieldValueWrapper values.
+  // We only accept FieldValueWrapper to help typing
+  //
+  // An example would be if a user requests "grouping" in an App Search API request. That will come back
+  // as "_group: {..}". It *should* be there in the Result so that a developer has it available to work
+  // with.
+  interface ResultT {
```
alexanderatallah: same here

packages/react-search-ui/react-search-ui.d.ts
```diff
@@ -0,0 +1,345 @@
...
+    // Props
+
+    className?: string;
+    label?: string;
+    sortOptions: SortOption[];
+    view?: renderFunc;
+
+    // State
+
+    sortDirection?: "asc" | "desc" | "";
```
alexanderatallah: realized that this could be `SortDirection | ""`. Also had a general question: what's the difference between an undefined sort direction and one equal to ""?

francoisprunier:
Hi,
I'm guessing the bot closing this PR is an error, can you confirm this is this still planned ?
Thanks

subsocialdev:
Hi guys, we need TypeScript support :)
TS is a must-have in medium to large web apps.

JasonStoltz:
> I'm guessing the bot closing this PR is an error, can you confirm this is this still planned ?

This is not currently planned. I would love to add support for this but I haven't been able to find the time.

JasonStoltz:
> TS is a must-have in medium to large web apps.

I hear you.

Issue #464, opened 2020-02-15
Cannot dynamically adjust Facets within a SearchProvider


alexanderatallah:
It looks like once a `SearchProvider` has been initialized, there's no way to adjust the `facets` part of the search `Context`. Am I right about this limitation?

This means that when we have React components dynamically change the aggregations needed in our Elasticsearch queries, we can't change the search `Context`'s `facets` and use them to build a request.

Here's the code I'd like to execute in place of your Elasticsearch example: https://github.com/elastic/search-ui/blob/master/examples/elasticsearch/src/buildRequest.js#L43

```
export const buildRequest = (state: Partial<Context>) => {
  const {
    current,
    filters,
    facets,
    resultsPerPage,
    searchTerm,
    sortDirection,
    sortField
  } = state

  const aggs = buildAggregations(facets)
...
```
But currently, since I start with no faceting, `facets` is always `undefined` unless a child component can re-configure the SearchProvider on the top level.

Best solution would be if we could have `Facet` components encapsulate the configuration of facets, changing the next query that's executed. Then, parent components don't have to worry about the data dependencies of their children.

JasonStoltz:
Generally speaking that is true, yes.

However, if you are using Elasticsearch as your back end this is not true.

Since there is no connector available for Elasticsearch, it is on you to write the logic to determine which facets or aggregations are fetched. That means you can apply whatever logic you wish here https://github.com/elastic/search-ui/blob/master/examples/elasticsearch/src/buildRequest.js#L43.

Does that make sense?

JasonStoltz:
@alexanderatallah A fix for this is available in 1.4.0-rc.0. If you get a chance to test it out, let me know how it works for you.

PR #471, opened 2020-03-30
Dynamic Search Configuration

JasonStoltz:
## Description

This allows Search Configuration to be changed after mounting SearchProvider. This unlocks uses cases like:

- Dynamically changing the `facets` available in a Search Experience based on user selections
- Dynamically changing `search_fields` based on users selection (Imagine an experience where you have a toggle that says something like "Search In" with selection of "title", "user name", etc.".
- Dynamically changing the `result_fields` that are displayed

This works for the configurations `autocompleteQuery` and `searchQuery` on the `SearchProvider` configuration.

```jsx
function App() {
  const [searchQuery, setSearchQuery] = useState({
    facets : /*List of facets*/
  });

  const newSearchQuery = {
    facets : /*List of NEW facets*/
  }
 
  return (
    <SearchProvider config={ searchQuery }>
       <button onClick={() => setSearchQuery(newSearchQuery)}></button>
       // ...
    </SearchProvider>
  );
}
```

latest

## List of changes

- Dynamic `autocompleteQuery`
- Dynamic `searchQuery`
- SearchProvider now accepts a driver as a parameter. Mostly for testing at this point.

## Associated Github Issues
Closes https://github.com/elastic/search-ui/issues/464
Unblocks: https://github.com/elastic/app-search-team/issues/861

### DIFF
--- a/packages/react-search-ui/src/SearchProvider.js
+++ b/packages/react-search-ui/src/SearchProvider.js
@@ -10,29 +10,40 @@ import defaultA11yMessages from "./A11yNotifications";
  * The SearchProvider primarily holds a reference to the SearchDriver and
  * exposes it to the rest of the application in a Context.
  */
-const SearchProvider = ({ children, config }) => {
-  const [driver, setDriver] = useState(null);
+const SearchProvider = ({ children, config = {}, driver }) => {
+  const [driverInstance, setDriverInstance] = useState(null);
 
   useEffect(() => {
     // This initialization is done inside of useEffect, because initializing the SearchDriver server side
     // will error out, since the driver depends on window. Placing the initialization inside of useEffect
     // assures that it won't attempt to initialize server side.
-    const newDriver = new SearchDriver({
-      ...config,
-      a11yNotificationMessages: {
-        ...defaultA11yMessages,
-        ...config.a11yNotificationMessages
-      }
-    });
-    setDriver(newDriver);
+    const currentDriver =
+      driver ||
+      new SearchDriver({
+        ...config,
+        a11yNotificationMessages: {
+          ...defaultA11yMessages,
+          ...config.a11yNotificationMessages
+        }
+      });
+    setDriverInstance(currentDriver);
+
     return () => {
-      newDriver.tearDown();
+      currentDriver.tearDown();
     };
   }, []);
 
+  // This effect allows users to dynamically update their searchQuery without re-mounting a SearchProvider,
+  // which would be destructive. An example of why this is useful is dynamically updating facets.
+  useEffect(() => {
+    if (driverInstance) {
+      driverInstance.setSearchQuery(config.searchQuery);
+    }
+  }, [config.searchQuery]);
+
   // Since driver is initialized in useEffect above, we are waiting
   // to render until the driver is available.
-  if (!driver) return null;
+  if (!driverInstance) return null;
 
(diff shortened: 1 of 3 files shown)


packages/react-search-ui/src/SearchProvider.js
```diff
@@ -10,29 +10,46 @@ import defaultA11yMessages from "./A11yNotifications";
  * The SearchProvider primarily holds a reference to the SearchDriver and
  * exposes it to the rest of the application in a Context.
  */
-const SearchProvider = ({ children, config }) => {
-  const [driver, setDriver] = useState(null);
+const SearchProvider = ({ children, config = {}, driver }) => {
```
JasonStoltz: I changed this to accept a driver as a parameter mostly so we can pass in a mock driver in tests.

packages/react-search-ui/src/SearchProvider.js
```diff
@@ -10,29 +10,46 @@ import defaultA11yMessages from "./A11yNotifications";
...
+          ...config.a11yNotificationMessages
+        }
+      });
+    setDriverInstance(currentDriver);
+
     return () => {
-      newDriver.tearDown();
+      currentDriver.tearDown();
     };
   }, []);
```
JasonStoltz: This effect runs once on initialization and tears itself down on unmount.

packages/react-search-ui/src/SearchProvider.js
```diff
@@ -10,29 +10,46 @@ import defaultA11yMessages from "./A11yNotifications";
...
 
+  // This effect allows users to dynamically update their searchQuery without re-mounting a SearchProvider,
+  // which would be destructive. An example of why this is useful is dynamically updating facets.
+  useEffect(() => {
```
JasonStoltz: This effect (and the next effect) are meant to only run after the initial mount. So this is basically equivalent to `componentDidUpdate`. The first time it runs on mount it will do nothing since `driverInstance` will be null.

packages/react-search-ui/src/SearchProvider.js
```diff
@@ -10,29 +10,46 @@ import defaultA11yMessages from "./A11yNotifications";
...
+    if (driverInstance) {
+      driverInstance.setSearchQuery(config.searchQuery);
+    }
+  }, [config.searchQuery]);
```
byronhulcher: Would you want to pass `driverInstance` into this too? I've read elsewhere that it's good form to include _all_ variables you reference in a hook here

JasonStoltz: Yeah that makes sense. I think for now it will be functionally equivalent but it would protect us from errors in the future.

JasonStoltz: I'm actually going to leave this as is, because there is a functional difference.

I ONLY want this to run when the searchQuery config is changed from the original value.

By adding `driverInstance`, I'm also running it whenever `driverInstance` changes.

When this component mounts, the first effect runs and ends up calling `setDriverInstance`, which calls this method on mount as well.

Luckily, I have a test for this so it broke when I attempted it 😄.

byronhulcher: Ha! Glad you found the issue so quick.  I'm going to look more into this for my own curiosity, sorry to send you on a goose chase.

Issue #468, opened 2020-03-26
Feature request: Config option to disable/bypass query cache


byronhulcher:
As far as I understand the default `SearchProvider` will cache the responses to queries based on the query's request.  I have a use case where I need to force Search UI to repeatedly poll a single query until it starts to see `meta.page.total_results > 0`.  I do not believe this is not possible given current options available to me in Search UI.

PR #472, opened 2020-03-30
Added AppSearchAPIConnector additional options

JasonStoltz:
## Description

Updated the App Search API Connector to accept additional options, which will be passed through to the API client.

## Associated Github Issues

Closes https://github.com/elastic/search-ui/issues/468

### DIFF
--- a/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
+++ b/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
@@ -53,7 +53,8 @@ class AppSearchAPIConnector {
     beforeAutocompleteResultsCall = (queryOptions, next) => next(queryOptions),
     beforeAutocompleteSuggestionsCall = (queryOptions, next) =>
       next(queryOptions),
-    endpointBase = ""
+    endpointBase = "",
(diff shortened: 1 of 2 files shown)


packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
```diff
@@ -53,7 +53,8 @@ class AppSearchAPIConnector {
     beforeAutocompleteResultsCall = (queryOptions, next) => next(queryOptions),
     beforeAutocompleteSuggestionsCall = (queryOptions, next) =>
       next(queryOptions),
-    endpointBase = ""
+    endpointBase = "",
+    ...rest
```
byronhulcher: This is a very clean approach BUT means that any other new parameters automatically become "breaking changes".  In this case _probably_ fine since we control the App Search API as well, but wanted to highlight that.

JasonStoltz: Good point. Hmmm, I think it's probably going to be fine also, but that isn't something I had considered. Thank you for mentioning it.

Issue #276, opened 2019-06-25
Cannot use hooks because of how containers call their views


FokkeZB:
The following:

```jsx
const MyResultView = () => {
  const someHook = useHook();

  return <div>My Result</div>;
};

<Result view={MyResultView} />
```

Errors with:
https://reactjs.org/warnings/invalid-hook-call-warning.html

I think this is because of how containers call their views:

https://github.com/elastic/search-ui/blob/c68f78f9f44680f8219c8aa1e2ed5442c56936af/packages/react-search-ui/src/containers/Result.js#L43-L50

When I change that to:

```jsx
    return <View
      className={className},
      result={result}
      key={`result-${result.id.raw}`}
      onClickLink={() => this.handleClickLink(result.id.raw)}
      titleField={titleField}
      urlField={urlField}
    />;
```

Everything works fine.

FokkeZB:
For others who run into this, you can workaround via:

```jsx
const MyResultView = () => {
  const someHook = useHook();

  return <div>My Result</div>;
};

const MyPatchedResultView = (props) => (
  <ResultView {...props} />
);

<Result view={MyPatchedResultView} />
```

constancecchen:
For what it's worth, I've just discovered I can reproduce this even without React hooks. For some odd reason the error messaging from React is different with vs. without React hooks.

### Without errors:

<img width="546" alt="" src="https://user-images.githubusercontent.com/549407/60372793-c6337400-99b2-11e9-8d35-c7283e08261e.png">

Note that **all** of the props being passed into Layout and Results are React functional components.

### With errors:

<img width="601" alt="" src="https://user-images.githubusercontent.com/549407/60372961-52459b80-99b3-11e9-816e-49d581880282.png">

<img width="416" alt="" src="https://user-images.githubusercontent.com/549407/60372966-5671b900-99b3-11e9-932b-cf1921719719.png">

<img width="1040" alt="" src="https://user-images.githubusercontent.com/549407/60372968-57a2e600-99b3-11e9-88e0-c707b99ac573.png">

The reason I wanted to try converting `Game` to `<Game />` was to pass a prop/method from the top level App to the child Game - something I likely can't do without the ability to pass it as an actual React component.

I also do think from a consistency standpoint that it makes sense to have `view` and `renderResult` take the same React component format that Layout's `bodyContent`, `sideContent` etc. all do.

Let me know if that makes sense at all and hopefully I'm not barking up the wrong tree here!

[...]

PR #476, opened 2020-04-01
This commit attempts to simplify views

JasonStoltz:
## Description

When users are creating custom views, they are instructed
to copy and paste the existing view from source and customize
it.

currently, there are a number of external imports that make the
copy/paste experience difficult.

This commit changes these views to all accept a "viewHelpers" prop,
which contain all view helpers for use. This means they can be
easily used without figuring out how to deal with imports.

This also had the side benefit for allowing us to delete duplicated
types throughout the two projects. I simply changed to the prop type
checks to be simpler in the view project and just use primitive
prop types.

For example, note the difference between the old view and the new view:

Old: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui-views/src/Result.js
New: https://github.com/JasonStoltz/search-ui/blob/simpler-views/packages/react-search-ui-views/src/Result.js

Note that the new view looks significantly less daunting to copy/paste and than customize.

## List of changes
- Officially deprecates the "Facets" component
- Replace all inline and imported functions with a new "viewHelpers" prop


packages/react-search-ui-views/src/Autocomplete.js
```diff
@@ -1,34 +1,8 @@
...
-  if (
-    autocompleteSuggestions[suggestionType] &&
-    autocompleteSuggestions[suggestionType].sectionTitle
-  ) {
-    return autocompleteSuggestions[suggestionType].sectionTitle;
-  }
-}
-
 function Autocomplete({
+  viewHelpers,
```
constancecchen, line 5: Just to check my high level understanding of this PR: why move this to a passed down prop instead of either 1. importing it or 2. using a hook?

I sort of get it if the goal is to expose `viewHelpers` to end-developers, but I also still hold that a prop isn't the most elegant way to do that - I think devs would prefer it to be a helper/hook they can choose to import only if they want it.

constancecchen: After rereading the examples in the PR description and mulling on it more, I see the argument for a passed prop - I'm still leaning a little towards thinking a hook is a bit nicer/more modern, but don't want to block this PR/release for that. 👍 from me if others are on board

JasonStoltz: Yeah, just to re-iterate, many people create custom views by copying the existing views from source, so like https://github.com/elastic/search-ui/blob/master/packages/react-search-ui-views/src/Result.js.

The following trips them up:
```
import { appendClassName, getUrlSanitizer } from "./view-helpers";
```

That's a relative import from within the project. When they use it externally, it's confusing if / how they can import those view helpers themselves.

Passing them as a prop is just a very low friction way of addressing this. They don't have to figure out how to import anything.

Your point about hooks is totally valid, though I think that would still require an import. We currently can't use hooks in views though 🤦 : https://github.com/elastic/search-ui/issues/276.

There may be a way though, through some build magic that we could change the import to something like:

```
import { appendClassName, getUrlSanitizer } from "@elastic/react-search-ui-views/view-helpers";
```

In such a way that it works internal to the project and external.

I'm not married to this approach so if there is clearly a more elegant approach I am open to it.

constancecchen: I'd definitely be +1 to a non-relative import like `import { appendClassName, getUrlSanitizer } from "@elastic/react-search-ui-views/view-helpers";` - that matches a lot of what other React libraries do as a best practice.

I guess the question is, would you rather merge this PR is as-is and change the prop/import behavior later in a major/breaking release, or would you be rather either delay the release or punt the PR to another release?

PR #481, opened 2020-04-14
Additional checks on Result component

JasonStoltz:
Added additional checks to the Result view component.

### DIFF
--- a/packages/react-search-ui-views/src/Result.js
+++ b/packages/react-search-ui-views/src/Result.js
@@ -1,7 +1,7 @@
 import PropTypes from "prop-types";
 import React from "react";
 
-import { appendClassName } from "./view-helpers";
+import { appendClassName, getUrlSanitizer } from "./view-helpers";
 import { isFieldValueWrapper } from "./types/FieldValueWrapper";
 
 function getFieldType(result, field, type) {
@@ -59,7 +59,7 @@ function Result({
 }) {
   const fields = getEscapedFields(result);
   const title = getEscapedField(result, titleField);
-  const url = getRaw(result, urlField);
+  const url = getUrlSanitizer(URL, location)(getRaw(result, urlField));
 
   return (
     <li className={appendClassName("sui-result", className)} {...rest}>
--- a/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
+++ b/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
@@ -221,7 +221,7 @@ exports[`renders correctly when there is a title and url 1`] = `
           "__html": "Title",
         }
       }
-      href="http://www.example.com/"
+      href="http://www.example.com"
       onClick={[Function]}
       rel="noopener noreferrer"
       target="_blank"
--- a/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
+++ b/packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
@@ -90,6 +90,60 @@ exports[`renders correctly when there is a URL 1`] = `
 </li>
 `;
 
+exports[`renders correctly when there is a malicious URL 1`] = `
+<li
+  className="sui-result"
+>
+  <div
+    className="sui-result__header"
+  />
+  <div
+    className="sui-result__body"
+  >
+    <ul
+      className="sui-result__details"
+    >
+      <li
+        key="field"
+      >
+        <span
+          className="sui-result__key"
+        >
+          field
+        </span>
+         
+        <span
+          className="sui-result__value"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": "value",
+            }
+          }
+        />
+      </li>
+      <li
+        key="url"
+      >
+        <span
+          className="sui-result__key"
+        >
+          url
+        </span>
+         
+        <span
+          className="sui-result__value"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": "javascript://test%0aalert(document.domain)",
+            }
+          }
+        />
+      </li>
+    </ul>
+  </div>
+</li>
+`;
+
 exports[`renders correctly when there is a title 1`] = `
 <li
   className="sui-result"
@@ -167,7 +221,7 @@ exports[`renders correctly when there is a title and url 1`] = `
           "__html": "Title",
         }
       }
-      href="http://www.example.com"
+      href="http://www.example.com/"
       onClick={[Function]}
       rel="noopener noreferrer"
       target="_blank"
--- a/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
+++ b/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
@@ -1,0 +1,95 @@
+import { getUrlSanitizer } from "../../view-helpers";
+import URL from "core-js-pure/features/url";
+
+describe("getUrlSanitizer", () => {
+  let url;
+  let currentLocation;
+
+  beforeEach(() => {
+    url = "";
+    currentLocation = "";
+  });
+
+  function subject() {
+    return getUrlSanitizer(URL, currentLocation)(url);
+  }
+
+  describe("when valid url with http", () => {
+    beforeEach(() => {
+      url = "http://www.example.com/";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("should allow it", () => {
+      expect(subject()).toEqual(url);
+    });
+  });
+
+  describe("when valid url with https", () => {
+    beforeEach(() => {
+      url = "https://www.example.com/";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("should allow it", () => {
+      expect(subject()).toEqual(url);
+    });
+  });
+
+  describe("when relative url", () => {
+    beforeEach(() => {
+      url = "/item/1234";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("should allow it", () => {
+      expect(subject()).toEqual("http://www.mysite.com/item/1234");
+    });
+  });
+
+  describe("when the protocol is javascript", () => {
+    beforeEach(() => {
+      url = "javascript://test%0aalert(document.domain)";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("should disallow it", () => {
+      expect(subject()).toEqual("");
+    });
+  });
+
+  describe("when the url is invalid", () => {
+    beforeEach(() => {
+      url = "<div>My bad URL</div>";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("treats it as a relative url, which should still be safe", () => {
+      expect(subject()).toEqual(
+        "http://www.mysite.com/%3Cdiv%3EMy%20bad%20URL%3C/div%3E"
+      );
+    });
+  });
+
+  describe("when the protocol is not whitelisted", () => {
+    beforeEach(() => {
+      url = "mailto://user@example.com";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("treats it as a relative url, which should still be safe", () => {
+      expect(subject()).toEqual("");
+    });
+  });
+
+  describe("when the url is protocol-less", () => {
+    beforeEach(() => {
+      url = "//www.example.com/";
+      currentLocation = "https://www.mysite.com";
+    });
+
+    it("uses the protocol from the current location", () => {
+      expect(subject()).toEqual("https://www.example.com/");
+    });
+  });
+});
--- a/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
+++ b/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
@@ -77,7 +77,7 @@ describe("getUrlSanitizer", () => {
       currentLocation = "http://www.mysite.com";
     });
 
-    it("treats it as a relative url, which should still be safe", () => {
+    it("should disallow it", () => {
       expect(subject()).toEqual("");
     });
   });
--- a/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
+++ b/packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
@@ -43,7 +43,7 @@ describe("getUrlSanitizer", () => {
     });
 
     it("should allow it", () => {
-      expect(subject()).toEqual("http://www.mysite.com/item/1234");
+      expect(subject()).toEqual(url);
     });
   });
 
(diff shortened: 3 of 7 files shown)


packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+  // This function is curried so that dependencies can be injected and don't need to be mocked in tests.
+  return url => {
+    if (!url) return url;
+
+    let parsedUrl;
+
+    try {
+      // Attempts to parse a URL as absolute
+      parsedUrl = new URLParser(url);
+      // eslint-disable-next-line no-empty
```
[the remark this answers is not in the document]
byronhulcher: Ah I see you need both blocks to check for absolute AND relative urls (good call).  I still think there's probably a way to write this w/o the disables.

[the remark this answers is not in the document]
JasonStoltz: That would change the behavior. I want it to try the absolute URL parsing, and if that fails, continue on to the second attempt using relative parsing.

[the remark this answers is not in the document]
JasonStoltz: I guess I could do that in the second catch if you think it would be better 🤔

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+    try {
+      // Attempts to parse a URL as relative
+      parsedUrl = new URLParser(url, currentLocation);
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
+
+    if (!parsedUrl) return "";
+
+    return VALID_PROTOCOLS.includes(parsedUrl.protocol)
+      ? parsedUrl.toString()
```
orhantoy, line 30: How about returning the original `url` here instead, just to avoid the URL being changed? I could only find that it gets changed in the case of `http://example.com` becoming `http://example.com/` (adds trailing slash) which is not a big deal.

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ *
+ * @param {URL} URLParser URL interface provided by browser https://developer.mozilla.org/en-US/docs/Web/API/URL
+ * @param {String} currentLocation String representation of the browser's current location
+ */
+export default function getUrlSanitizer(URLParser, currentLocation) {
+  // This function is curried so that dependencies can be injected and don't need to be mocked in tests.
+  return url => {
+    if (!url) return url;
```
orhantoy, line 11: This check isn't strictly needed because if the parsing later on fails, we'll return an empty string (which isn't exactly the same as this).

JasonStoltz: I put this in because if this is not here, it attempts to parse an empty string as a relative url, so it ends up returning `http://example.com/undefined` or `http://example.com/`, etc

packages/react-search-ui-views/src/Result.js
```diff
@@ -59,7 +59,7 @@ function Result({
 }) {
   const fields = getEscapedFields(result);
   const title = getEscapedField(result, titleField);
-  const url = getRaw(result, urlField);
+  const url = getUrlSanitizer(URL, location)(getRaw(result, urlField));
```
constancecchen, line 62: Sorry, this is me being real slow in the morning, but why are we currying a native URL function here but an imported core-js URL function in our tests?

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
JasonStoltz: That's a good point. Let me do that, it would also let me remove that additional check you mentioned in your other comment.

packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
```diff
@@ -90,6 +90,60 @@ exports[`renders correctly when there is a URL 1`] = `
...
+          url
+        </span>
+         
+        <span
+          className="sui-result__value"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": "javascript://test%0aalert(document.domain)",
+            }
+          }
```
constancecchen, line 139: Jest snapshots never fail to confuse me sometimes. no idea if it's just me, but `dangerouslySetInnerHTML` here seems to be fairly counterintuitive, shouldn't this just be plain text within a span? 🤷 

that being said if this is of course auto generated code, nothing to do here, just me yelling at clouds

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+
+    let parsedUrl;
+
+    try {
+      // Attempts to parse a URL as absolute
+      parsedUrl = new URLParser(url);
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
```
constancecchen, line 19: Just to make sure I'm on the same page, an empty catch swallows any error without failing/stopping correct? So we're basically just continuing if this fails?

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
orhantoy: Oh right, I didn't consider the relative case :v:

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
```diff
@@ -0,0 +1,29 @@
...
+
+    try {
+      // Attempts to parse a URL as relative
+      parsedUrl = new URLParser(url, currentLocation);
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
+
+    if (!parsedUrl) return "";
```
constancecchen, line 25: Is this if check here necessary? a falsey parsedUrl will still fail the ternary below and return a `""`, right?

packages/react-search-ui-views/src/Result.js
JasonStoltz: Because URL is available in the browser, but not in the node environment that jest runs

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
JasonStoltz: Yes

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
orhantoy: Just saw https://github.com/elastic/search-ui/pull/481#discussion_r408185113 and yes, that change would also make this one possible.

packages/react-search-ui-views/src/__tests__/__snapshots__/Result.test.js.snap
JasonStoltz: @constancecchen After running our new code, `url` will be empty if a malicious url is attempted. 

In that case, this line of code runs: https://github.com/elastic/search-ui/blob/master/packages/react-search-ui-views/src/Result.js#L67

That line of code prints the title, which may or may not have snippets, so it uses `dangerouslySetInnerHTML`.

The title is html escaped [here](https://github.com/elastic/search-ui/blob/master/packages/react-search-ui-views/src/Result.js#L61), which should make that safe.

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
constancecchen: If this is generally for dev readability however feel free to ignore me and leave this in!

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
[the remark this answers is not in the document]
constancecchen: Maybe I'm not understanding how try/catch works, but:

> I want it to try the absolute URL parsing, and if that fails, continue on to the second attempt using relative parsing.

Regardless of whether it fails or not, we're continuing on to relative parsing because we're not checking for `if parsedUrl` or anything in the 2nd try block.

Do we even need the first try/catch block in that case? What happens if we remove it?

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
JasonStoltz: You're right, it's no longer necessary. I'd just need to update the ternary below to ensure parsedUrl is not undefined before calling `parsedUrl.protocol`.

```
return VALID_PROTOCOLS.includes((parsedUrl || {}).protocol) ? url : "";
```

Or maybe just change the initial assignment of parsedURL

```
let parsedUrl = {};
```

JasonStoltz: https://github.com/elastic/search-ui/pull/481/commits/c4c9bc75ecddb9b3ac26275bbd8bd9feae3e996a

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
JasonStoltz: https://github.com/elastic/search-ui/pull/481/commits/57655049d66b550aa135add7d8088571d37f4442

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
JasonStoltz: https://github.com/elastic/search-ui/pull/481/commits/57655049d66b550aa135add7d8088571d37f4442

packages/react-search-ui-views/src/Result.js
constancecchen: Ahhh - hm, is this not the browser equivalent of URL()? https://nodejs.org/api/url.html

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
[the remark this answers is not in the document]
JasonStoltz: If the path is relative it will throw because it is an invalid URL

```
> const url = new URL('/relative-path')
> VM222:1 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at <anonymous>:1:13
```

I catch this so that doesn't get passed on to the user, I'd rather just attempt to parse this as a relative URL.

Does that make sense?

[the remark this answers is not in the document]
JasonStoltz: Sorry this code is so confusing, I could separate those out into separate function if it helps readability.

packages/react-search-ui-views/src/Result.js
constancecchen: Oh wait, right, the node env specific to Jest 🤦‍♀️ 

I believe there is a specific flag or config you can use at the top of the test to tell Jest to run it as a node env (instead of jest's default jsdom env), but not sure if you want to go back and look at it now rather than focusing on getting this out and shipped.

After some thought I think my personal preference would have been mocking URL() in Jest rather than currying it (which for some reason just makes me a little nervous that we're potentially not in full control of the URL parser/is another possible vector for attack), but that's just me being paranoid and splitting hairs :)

packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
```diff
@@ -0,0 +1,95 @@
...
+      url = "/item/1234";
+      currentLocation = "http://www.mysite.com";
+    });
+
+    it("should allow it", () => {
+      expect(subject()).toEqual("http://www.mysite.com/item/1234");
+    });
+  });
+
+  describe("when the protocol is javascript", () => {
```
richkuz, line 50: Just because I'm paranoid, can we also include a test for `" javascript://test%0aalert(document.domain)"` (with a space)? Seems like [a non-obvious case](https://medium.com/javascript-security/avoiding-xss-in-react-is-still-hard-d2b5c7ad9412) that could trip up some folks when a location isn't specified.

packages/react-search-ui-views/src/view-helpers/getUrlSanitizer.js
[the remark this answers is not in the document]
constancecchen: Sorry, my question was more along the lines of "can we just do `new URLParser(url, currentLocation)` to start with and skip the first `new URLParser(url)`?" - I def agree we shouldn't skip the 2nd parser with currentLocation.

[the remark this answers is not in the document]
constancecchen: From what I can tell (from running the test for this file) nothing seems to break when removing lines 13-17.

I also tested this in the browser:
```
new URL('/foo/bar', 'https://github.com')
```
```
new URL('http://foo.bar', 'https://github.com')
```

Both seem to work just fine.

packages/react-search-ui-views/src/__tests__/view-helpers/getUrlSanitizer.test.js
JasonStoltz: https://github.com/elastic/search-ui/pull/481/commits/ac3b5bc88ce4f7bc188c332e11d11647f76b76ac

packages/react-search-ui-views/src/Result.js
JasonStoltz: Oh I'm sorry, I didn't realize there was a URL class available in node. I wonder if it is consistent with the browser's API.

JasonStoltz: Any specific thought around how it could be used as an attack vector? We could probably change this.

richkuz:
I approve.

JasonStoltz:
@constancecchen I noticed it yesterday but haven't had a chance to fix it. We don't lock in our package dependencies so I have a feeling a change in an underlying dependency broke CI.

I can address this today. I'm going to force merge this PR though, since the CI failures seem to be consistent with the CI failures on master.

packages/react-search-ui-views/src/Result.js
constancecchen: Nah it totes can't, sorry for the unclear verbal diarrhea. What I meant was more that whenever I see something as an arg, mentally I tend to think "this is dynamic and changes" and also sometimes "the user is passing this in", vs. a regular import/constant which definitely reads to me as "this is defined by only us".

If we had infinite time I'd like to come back and maybe look at converting this to mocking URL() in jest instead of passing a curried function, but that can be tech debt for another day!

JasonStoltz:
Awesome, thank again for the help all!

packages/react-search-ui-views/src/Result.js
JasonStoltz: Makes total sense. I'll still need to port this forward to master, so I could try to clean it up then.

PR #496, opened 2020-05-27
Remove unnecessary array brackets invoking parsing errors in BuildReq…

Gldud1118:
## Description
I have tested your example project. When I run the project and use filters to filter data, it throws 400 bad request error.

Error messages said "failed to parse field [something]".

It turned out that there's something wrong in BuildRequestFilter.js. When you write queries for "request body parameters", you don't have to add "[]" to make an array because map function return an array. Array inside array invoked parsing errors, I think.

After I remove all unnessary array brackets and tested it, everything worked out well.

## List of changes
Remove unnessary array brackets in BuildRequestFilter.js

### DIFF
--- a/examples/elasticsearch/src/buildRequestFilter.js
+++ b/examples/elasticsearch/src/buildRequestFilter.js
@@ -15,22 +15,18 @@ function getTermFilter(filter) {
   if (filter.type === "any") {
     return {
       bool: {
-        should: [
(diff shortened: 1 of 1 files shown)


JasonStoltz:
Oh no! I just fixed this in a separate PR, I didn't realize you had a fix here already.

Apologies for not getting to this sooner. But for now, this issue should be resolved so I am closing this PR.

PR #521, opened 2020-10-26
Fix Autocomplete handler error

### LINKED ISSUE #516: Uncaught TypeError: Cannot read property 'suggestion' of null
Search ui is throwing above error on pressing esc button.

Steps to reproduce :

1. Enter text for search and select search text from suggestions.
2. Press esc key to clear input it will throw error  as - "SearchBox.js?9b6a:117 Uncaught TypeError: Cannot read property 'suggestion' of null"

JasonStoltz:
There are cases where our autocomplete handler will be triggered with no selection. Specifically, when the esc key is pressed on the input after making a selection. This is probably a quirk of [downshift](https://github.com/downshift-js/downshift). Since we likely can't prevent this from happening, the fix in this case is to simply put a check in our handler to exit early if no selection is available.


constancecchen (approved):

constancecchen:
In theory, I'm slightly worried about this fix since we can't reliably repro the original bug & the early return is a huge change from everything that follows. In practicality I think it'll be fine, but would suggest keeping an eye on this post-release.

JasonStoltz:
I was able to reproduce the originally reported bug. I am just uncertain of why the event was firing when it was.

constancecchen:
Ahh gotcha!

JasonStoltz:
@constancecchen You did not, it was totally unclear.

JasonStoltz:
I updated the description.

PR #568, opened 2021-06-07
[App Search Connector] Made searchKey optional

JasonStoltz:
## Description

We have certain use cases, internal to App Search, where we prefer to rely on App Search's session based authentication, rather than key based authentication.

To enable this, I simply made the `searchKey` parameter optional. If it is not present, we will not send an authorization header to the browser, which implicitly prompts App Search to authenticate the request based on session.

## Associated Github Issues

https://github.com/elastic/app-search-javascript/pull/30


packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.js
```diff
@@ -105,6 +105,17 @@ describe("AppSearchAPIConnector", () => {
     expect(connector).toBeInstanceOf(AppSearchAPIConnector);
   });
 
+  // The use case for this is mostly internal to Elastic, where we rely on the logged in user session (via cookies) to authenticate
```
constancecchen, line 108: Super appreciate this contextual comment!

constancecchen (approved):
LGTM

Issue #539, opened 2021-02-09
Error: Unable to import `../package.json`


adrhumphreys:
I'm not sure what is causing this, my webpack build chain functions normally, but when you then load the app in the browser you get the error. Perhaps it's an issue with using relative paths and then in dev mode it producing a relative path that the browser is trying to open.

By updating https://github.com/elastic/search-ui/blob/3cfe66660a0b976443122160a4ec28a7b8a1ca5e/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js#L2 to:
```
const version = '1.5.0';
```

I was able to get around the issue as a very temporary patch

Edit:
Also tried to update it to: `import { version } from "@elastic/search-ui-app-search-connector/package.json";` which still produced the same issue:
```
Uncaught Error: Cannot find module '@elastic/search-ui-app-search-connector/package.json'
    at webpackMissingModule (package.json:1)
    at eval (package.json:1)
    at Object../node_modules/@elastic/search-ui-app-search-connector/package.json (bootstrap:83)
    at __webpack_require__ (bootstrap:19)
    at eval (AppSearchAPIConnector.js:16)
    at Module../node_modules/@elastic/search-ui-app-search-connector/es/AppSearchAPIConnector.js (bootstrap:83)
    at __webpack_require__ (bootstrap:19)
    at eval (index.js:2)
    at Module../node_modules/@elastic/search-ui-app-search-connector/es/index.js (bootstrap:83)
    at __webpack_require__ (bootstrap:19)
```

Edit 2:
Looks like the fool is me, babel was trying to read it and, then not being able too, so it would just exclude it from the build.
Updating my babel config to ignore `*/**/package.json` fixed the issue for me

simofacc:
Hi,

This issue is probably related to webpack 5 named exports https://webpack.js.org/migrate/5/#using-named-exports-from-json-modules.

ctestamarck:
NextJS with Webpack5 shows this warning: `Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)`

Was able to patch per Webpack documentation: 

`diff --git a/node_modules/@elastic/search-ui-app-search-connector/es/AppSearchAPIConnector.js b/node_modules/@elastic/search-ui-app-search-connector/es/AppSearchAPIConnector.js
index d6c5e74..cd11fe2 100644
--- a/node_modules/@elastic/search-ui-app-search-connector/es/AppSearchAPIConnector.js
[...]

PR #572, opened 2021-06-29
Update package.json imports to use default export for Webpack 5

Pearce-Ropion:
## Description
As per Webpack 5's migration instructions, webpack is removing support for importing named properties from a default export. As such, all instances of importing individual keys from the package.json should be updated to import the default export.
https://webpack.js.org/migrate/5/#using-named-exports-from-json-modules

For example
```js
// old
import { version } from "./package.json";
const foo = version; // example

// new 
import packageJson from "./package.json";
const foo = packageJson.version; // example
```

## List of changes
Updated all package.json imports to import the full package.json and then use properties as you normally would from an object.

As this is only a minor semantics change, I tested by running `npm run test`.

## Associated Github Issues
#539

### DIFF
--- a/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
+++ b/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
@@ -1,5 +1,5 @@
 import * as ElasticAppSearch from "@elastic/app-search-javascript";
-import { version } from "../package.json";
+import packageJson from "../package.json";
 
 import { adaptResponse } from "./responseAdapter";
 import { adaptRequest } from "./requestAdapters";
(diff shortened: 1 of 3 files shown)


JasonStoltz:
@Pearce-Ropion Looks great, can you sign the CLA please?

Pearce-Ropion:
Hmm, I thought I already did.

Pearce-Ropion:
Ok, definitely signed this time

JasonStoltz:
@Pearce-Ropion Can you check out https://github.com/elastic/search-ui/pull/573? I'd like to merge that as an alternative to this PR.

Pearce-Ropion:
> @Pearce-Ropion Can you check out #573? I'd like to merge that as an alternative to this PR.

I think #573  would be a be better solution as it removes the need for the version in the first place. I will close this now.

PR #573, opened 2021-06-30
Removed 'swiftype_' headers

### LINKED ISSUE #539: Error: Unable to import `../package.json`

JasonStoltz:
## Description

This PR removes the 'swiftype_' prefixed headers on API calls for the App Search and Site Search connectors.

These were originally introduced when App Search and Site Search were purely SaaS products, for our own analytical purpose. These are no longer used or needed by us (Elastic).

Since this causes complications in the build as seen in #539, I'd rather just remove them entirely.

Since these headers were for internal use, I don't consider this a breaking change.

## List of changes

Removed `x-swiftype-integration` and `x-swiftype-integration-version` headers.

## Associated Github Issues

Fixes #539
Alternative to #572

### DIFF
--- a/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
+++ b/packages/search-ui-app-search-connector/src/AppSearchAPIConnector.js
@@ -1,5 +1,4 @@
 import * as ElasticAppSearch from "@elastic/app-search-javascript";
-import { version } from "../package.json";
 
 import { adaptResponse } from "./responseAdapter";
 import { adaptRequest } from "./requestAdapters";
@@ -67,10 +66,6 @@ class AppSearchAPIConnector {
       ...(hostIdentifier && { hostIdentifier: hostIdentifier }),
       apiKey: searchKey,
       engineName: engineName,
-      additionalHeaders: {
-        "x-swiftype-integration": "search-ui",
-        "x-swiftype-integration-version": version
-      },
       ...rest
     });
     this.beforeSearchCall = beforeSearchCall;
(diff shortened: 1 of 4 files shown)


constancecchen (approved):
Nice catch!

Issue #552, opened 2021-04-16
Enable sorting on multiple fields


DeliDro:
I would like to be able to order my results with a sort on multiple fields such as: 
```
{
  name : "multiple_sort",  
  sortOptions: [
    {
      name: "field_a",
      value: "field_a",
      direction: "desc"
    },
    {
      name: "field_a",
      value: "field_a",
      direction: "desc"
     }
  ]
}
```

I know that the API allows it as mentionned [here](https://www.elastic.co/guide/en/app-search/current/sort.html#sort-multi-field-sorting).

JasonStoltz:
This is a great suggestion, thanks.

DeliDro:
Cool, can't wait

dgastudio:
any update on this? 

for now, i have solved this issue by using the following code

```
function buildSort(sortDirection, sortField) {

  if (sortDirection && sortField) {
    let sort = sortField.split('|')
    if (Array.isArray(sort) && sort.length > 0) {
      var sorts = [];
      sort.map(i => (
          sorts.push({[`${i}`]: 'ASC'})
      ))
      return sorts
    }
    else {
      return [
        {
          [`${sortField}`]: sortDirection,
        },
      ];
    }

  }
}

```

it works fine, but the only issue is with the sort field in url

JasonStoltz:
Unfortunately no, I haven't been able to work on this.

JasonStoltz:
Just noting here for inspiration on how to implement this feature: https://www.searchkit.co/docs/reference/schema#sorting

ckeboss:
@JasonStoltz I will take a look at implementing this API. Should be backwards compatible if we keep the `sortDirection` and `sortField`.

JasonStoltz:
@ckeboss Looks like you've seen the PR from @mtlewandowsk89. It's a great start, I dropped some feedback, if that can be addressed then I'll merge the PR and drop a new version.

JasonStoltz:
Closed by https://github.com/elastic/search-ui/pull/592

Released in 1.8.0.

PR #586, opened 2021-09-21
Support multiple sort options

mtlewandowsk89:
Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

These changes give the user the ability to sort by multiple different properties. The second and any additional properties will be used as tie breakers.

Allow search-ui to have multiple sort parameters.

https://github.com/elastic/search-ui/issues/552


mtlewandowsk89:
Let me know if this circleci error is something on my end I need to fix or not. It said something about it being caused by the CI environment potentially.

examples/sandbox/src/App.js
```diff
@@ -33,6 +33,28 @@ const SORT_OPTIONS = [
...
+    value: "states",
+    direction: "asc"
+  },
+  {
+    name: "State -> Title",
+    value: [
+      { "states" : "asc" },
+      { "title" : "asc" }
+    ],
+    direction: ""
```
ckeboss, line 48: This seems redundant if the direction is already passed in the `value`

packages/react-search-ui/src/containers/Sorting.js
```diff
@@ -2,24 +2,45 @@ import PropTypes from "prop-types";
...
-  );
+function findSortOption(sortOptions, sortData) {
+  let value;
+  let direction;
+  if (Array.isArray(sortData)) {
+    value = sortData;
+
+    // We don't care about option.direction in this case since direction is part of the value array.
+    return sortOptions.find(
+      option => JSON.stringify(option.value) === JSON.stringify(value)
```
ckeboss, line 17: Can you explain the reasoning behind stringifying these objects and then comparing the strings?

ckeboss (commented):
It also looks like the selected option is not selecting the correct value from the url state

packages/react-search-ui/src/containers/Sorting.js
mtlewandowsk89: It didn't work without JSON.stringify. The find method was returning undefined until I added the JSON.stringify.

mtlewandowsk89:
"It also looks like the selected option is not selecting the correct value from the url state". I will look into this. I didn't notice that at first.

JasonStoltz:
Thank you both!

martynball:
@ckeboss Awesome thanks!

Issue #557, opened 2021-05-03
dependency resolution error with React 17


tarun:
I just created a new `create-react-app` application and tried to add @elsatic/search-ui to it - which resulted in module resolution errors - as the app used use React 17 vs. @elastic/search-ui dependencies reference react 16.

it would be nice to have a new version of @elastic/react-search-ui with React 17 updated dependencies.

```
npx create-react-app test-search-ui --use-npm --template typescript
```

```
npm install --save @elastic/react-search-ui @elastic/search-ui-app-search-connector

.....

npm ERR! node_modules/react
npm ERR!   react@"^17.0.2" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0" from @elastic/react-search-ui@1.5.1
npm ERR! node_modules/@elastic/react-search-ui
npm ERR!   @elastic/react-search-ui@"*" from the root project
```

error log attached: [npm-error.txt](https://github.com/elastic/search-ui/files/6416199/npm-error.txt)

tarun:
I am trying this out on a fork now https://github.com/elastic/search-ui/compare/master...tarun:react-17
this would be a semver major change and need more testing.


one test failure to be looked into
```
@elastic/react-search-ui: FAIL src/__tests__/SearchProvider.test.js (33.849s)
@elastic/react-search-ui:   ● SearchProvider › will clean up searchDriver on unmount
@elastic/react-search-ui:     expect(jest.fn()).toHaveBeenCalled()
@elastic/react-search-ui:     Expected number of calls: >= 1
@elastic/react-search-ui:     Received number of calls:    0
@elastic/react-search-ui:       32 |
@elastic/react-search-ui:       33 |     wrapper.unmount();
@elastic/react-search-ui:     > 34 |     expect(driver.tearDown).toHaveBeenCalled();
@elastic/react-search-ui:          |                             ^
@elastic/react-search-ui:       35 |   });
@elastic/react-search-ui:       36 |
@elastic/react-search-ui:       37 |   it("will update searchDriver when searchQuery config changes", () => {
@elastic/react-search-ui:       at Object.<anonymous> (src/__tests__/SearchProvider.test.js:34:29)
```

JasonStoltz:
Thank you for reporting this! I'll see if we can get this updated.

JasonStoltz:
Just wanted to bump this to the top. Taking a look at it now.

JasonStoltz:
Still relevant

[...]

PR #605, opened 2022-01-14
#557 update peerDeps and tests to be compat with react 17

joemcelroy:
## Description
Small changes to update compatibility with react 17. 
- update peerDeps for packages which will depend on react & react-dom
- update enzyme adapter to use the unofficial react-17 adapter (enzyme doesn't yet support 17)
- update the example apps to use react 17

## Associated Github Issues
#557


yakhinvadim (approved):
LGTM 👍 
All tests are passing for me locally and the sandbox is running without issues.

PR #628, opened 2022-02-08
[WIP #612 app-search-connector TypeScript conversion

yakhinvadim:
WIP


packages/react-search-ui/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.9.0",
   "description": "A React library for building search experiences",
   "license": "Apache-2.0",
-  "main": "lib",
+  "main": "lib/cjs",
```
JasonStoltz, line 6: It seems like maybe this is something @joemcelroy missed in his react-search-ui PR?

In which case I think this should probably be:
```
  "main": "lib/cjs",
  "module": "lib/esm",
  "types": "./lib/esm/index.d.ts",
```

yakhinvadim: Decided to revert this change completely to avoid conflicts with Joe's PR.

packages/search-ui-app-search-connector/src/AppSearchAPIConnector.ts
```diff
@@ -3,6 +3,39 @@ import * as ElasticAppSearch from "@elastic/app-search-javascript";
...
+  documentId: string;
+  requestId: string;
+  tags: string[];
+}
+
+export type SearchQueryHook = (
+  queryOptions: QueryConfig,
+  next: (newQueryOptions: any) => any
+) => any;
+export type SuggestionsQueryHook = (
```
joemcelroy, line 35: is the return type here void?

packages/search-ui-app-search-connector/src/requestAdapters.ts
```diff
@@ -1,5 +1,21 @@
...
+  SortOption,
+  FilterValue,
+  FilterValueRange
+} from "@elastic/search-ui";
+
+// Using type predicates https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
+function isFilterValueRange(
+  filterValue: FilterValue
+): filterValue is FilterValueRange {
+  return (filterValue as FilterValueRange).name !== undefined;
```
joemcelroy, line 14: nice!

packages/search-ui/src/Events.ts
```diff
@@ -58,7 +58,7 @@ class Events {
   // eslint-disable-next-line  @typescript-eslint/no-explicit-any
   public autocomplete: (
     query: { searchTerm: string },
-    queryConfig: QueryConfig
+    queryConfig: AutocompleteQuery
```
joemcelroy, line 61: nice!

packages/search-ui/src/types/index.ts
```diff
@@ -50,9 +50,10 @@ export type RequestState = {
   filters?: Filter[];
   resultsPerPage?: number;
   searchTerm?: string;
-  sortDirection?: SortDirection;
-  sortField?: string;
-  sortList?: SortOption[];
+  sortDirection?: SortDirection; // Deprecated
+  sortField?: string; // Deprecated
+  sortList?: SortOption[]; // Maybe deprecated? Not in docs
+  sort?: SortOption[];
```
joemcelroy, line 56: yeh what was obvious whilst i was doing sorting types, it was a mixture of old and new. When we mature our TS conversion over next coming weeks, definitely need to zoom into sorting options and document them well.

joemcelroy (approved):

joemcelroy:
nice work! LGTM 👍

packages/search-ui-app-search-connector/src/AppSearchAPIConnector.ts
yakhinvadim: Nice catch! I updated return type of hooks in this commit: https://github.com/elastic/search-ui/pull/628/commits/2e24a3f4b58d2abfd3a95f0d62a6701b8d5a68df

I had to still keep `any` for the return type of `next`, because we're returning an adapted response there:

```
return this.beforeSearchCall(options, async (newOptions) => {
      const response = await this.client.search(query, newOptions);
      return adaptResponse(response, buildResponseAdapterOptions(queryConfig));
    });
```

I don't know the exact type since it's partially coming from a javascript client that doesn't have types yet, thus I have to keep `any`.

yakhinvadim: Nice catch! I went back and forth on this a bit (you can see in the last commits) but decided to keep `any` everywhere. Some `void` types are causing errors in other places and I'm not confident in assigning more specific types. Keeping it as-is for now.

Issue #626, opened 2022-02-07
Add example app to mono-repo workspace


joemcelroy:
- Add the example app to be part of the mono-repo workspace so its dependencies are hoisted to the root of the project and the mono-repo packages are symlinked to the app.
- Add a resolution for all packages where there cannot be more than one version. React and React-dom are the key dependencies where resolutions need to be added
- See example PR which demonstrate changes https://github.com/searchkit/searchkit/commit/ec874e110b766203b4c18a0cbba53c8fc9423390

PR #635, opened 2022-02-11
#626 yarn workspaces

joemcelroy:
**Problem**
- Individual packages which depend on react (react-search-ui & react-search-ui-views) had their own copy of react in their respective `node_modules` folder. This caused a conflict when these packages were symlinked to an example app which also had its own version of React / ReactDOM. React throws an exception when it detects multiple versions of react.

**Solution**
- Use yarn workspaces to hoist all similar dependencies across packages and example apps into the root, thus creating only one copy of particular libraries
- Use resolutions to make sure requests for different library versions requested wont create multiple copies. Used for react & react-dom.

**Changes that Devs need to do**
- You must now use yarn for dependency installs. 

https://user-images.githubusercontent.com/49480/153589788-3526bcfd-f33e-4a47-999c-83c3dc09d211.mov


lerna.json
```diff
@@ -4,13 +4,17 @@
     "packages/*"
   ],
   "version": "1.10.0-rc.0",
+  "useWorkspaces": true,
```
joemcelroy, line 7: more information on workspaces, see https://blog.kintone.io/entry/2020/07/14/103322. Other companies who use workspaces are vercel/next (which have a similar problem to us with React examples + packages)

package.json
```diff
@@ -2,8 +2,11 @@
   "name": "root",
   "private": true,
   "license": "Apache-2.0",
+  "workspaces": [
+    "packages/*",
+    "examples/*"
+  ],
   "scripts": {
-    "postinstall": "lerna bootstrap --no-ci",
```
joemcelroy, line 6: removed postInstall for lerna bootstrap as no longer needed with lerna workspaces.

packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
```diff
@@ -1,5 +1,5 @@
 @include block("paging") {
-  @import "../../node_modules/rc-pagination/dist/rc-pagination.min.css";
+  @import "../../../../node_modules/rc-pagination/dist/rc-pagination.min.css";
```
joemcelroy, line 2: short term fix
long term fix - we need a better solution for this...

Issue #613, opened 2022-01-18
Typescript: search-ui-site-search-connector


joemcelroy:
- Update code to typescript
- Update tests to typescript
- configure the package to use typescript for building commonjs and esm outputs

PR #637, opened 2022-02-14
#613 site-search-connector TypeScript conversion 

yakhinvadim:
This PR converts site-search-connector to TypeScript.

There were several places where tests and code were in conflict with the docs, I marked those with TODO and usually applied type "any".


packages/search-ui-app-search-connector/package.json
```diff
@@ -43,6 +42,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
+    "@elastic/search-ui": "^1.9.0",
```
yakhinvadim, line 45: Moved search-ui core package from devDependencies to dependencies, as I need to import a helper from there.

packages/search-ui-site-search-connector/src/request.ts
```diff
@@ -26,7 +31,10 @@ export default async function request(engineKey, method, path, params) {
   if (response.status >= 200 && response.status < 300) {
     return json;
   } else {
-    const message = json && json.error ? json.error : response.status;
-    throw new Error(message);
+    const message =
+      json && json.errors && Object.entries(json.errors).length > 0
+        ? JSON.stringify(json.errors)
+        : response.status;
+    throw new Error(`${message}`);
```
yakhinvadim, line 38: Had to change the code slightly here, since based on docs API returns "errors" field, not "error" field.

packages/search-ui-site-search-connector/bin/build-js
```diff
@@ -1,5 +1,7 @@
-# Builds Common JS JavaScript
-babel src --out-dir lib --ignore '**/*.test.js'
+# clear directory
+rm -rf ./lib
```
joemcelroy, line 2: you can take out the rm -rf as its been removed by rimraf in the NPM scripts

packages/search-ui-site-search-connector/bin/watch-js
```diff
@@ -1,2 +1,3 @@
-babel src --out-dir lib --ignore '**/*.test.js' --watch &
-BABEL_MODULES=false babel src --out-dir es --ignore '**/*.test.js' --watch &
+rm -rf ./lib
```
joemcelroy, line 1: same here

joemcelroy:
One thing (which can be in another PR too) is having an interface for APIConnector so when a developer decides to build a new APIConnector can implement from.

packages/search-ui-site-search-connector/bin/build-js
yakhinvadim: Nice catch! https://github.com/elastic/search-ui/pull/637/commits/386a32437aa3eceb3047ea4d93e9623030ccf347

PR #639, opened 2022-02-15
Release: v1.10.0-rc.1

yakhinvadim:


joemcelroy (changes requested):
I would also bump the versions for example apps to rc1 too :)

PR #650, opened 2022-03-02
Workplace search connector

yakhinvadim:
The first commit is a copy-paste of all App Search connector code. It may be easier to review further diff commit-by-commit.

This PR doesn't include docs update and Click API (Analytics event API in WS).


examples/sandbox/src/App.js
```diff
@@ -85,6 +86,13 @@ if (process.env.REACT_APP_SOURCE === "SITE_SEARCH") {
       process.env.REACT_SITE_SEARCH_ENGINE_KEY || "Z43R5U3HiDsDgpKawZkA",
     documentType: process.env.REACT_SITE_SEARCH_ENGINE_NAME || "national-parks"
   });
+} else if (process.env.REACT_APP_SOURCE === "WORKPLACE_SEARCH") {
+  connector = new WorkplaceSearchAPIConnector({
+    kibanaBase: "https://search-ui-sandbox.kb.us-central1.gcp.cloud.es.io:9243",
+    enterpriseSearchBase: "https://search-ui-sandbox.ent.us-central1.gcp.cloud.es.io",
+    redirectUri: "http://localhost:3000",
+    clientId: "8e495e40fc1e6acf515e557e534de39d4f727f7f60a3afed24a99ce3a6607c1e",
```
JasonStoltz: We should replace these with environment variables like we do for App Search, and possibly not provide defaults. It might not be a good idea to expose actual cloud details here.

examples/sandbox/src/App.js
```diff
@@ -217,87 +237,129 @@ const config = {
...
-                        filterType="any"
-                      />
-                      <Facet
-                        field="location"
-                        label="Distance"
-                        filterType="any"
+    <EuiProvider colorMode="light">
+      <SearchProvider config={config}>
+        <WithSearch
+          mapContextToProps={({ wasSearched, authorizeUrl, isLoggedIn }) => ({
```
JasonStoltz, line 243: I don't think we should add `isLoggedIn` and `authorizeUrl` to the Search UI state. Since it's specific to the connector, we should probably figure out some other way to pass this information around.

For `authorizeUrl`, you could likely do something like:
```
                        <EuiModalFooter>
                          <EuiButton href={connector.state.authorizeUrl} fill fullWidth>
                            Log in
                          </EuiButton>
                        </EuiModalFooter>
```

You could do something similar for the isLoggedIn state, but let me ponder what that might look like.

JasonStoltz: Maybe something like...

```jsx
const [isLoggedIn, setIsLoggedIn] = useState(connector.state.isLoggedIn);

  useEffect(() => {
    connector.onLoggedInChanged((newIsLoggedIn) => {
      setIsLoggedIn(newIsLoggedIn);
    });
  }, []);
```

examples/sandbox/src/App.js
yakhinvadim: Added env variables here: https://github.com/elastic/search-ui/pull/650/commits/55f8f89e2f9af6e1fa597646bcb01245ac6eceb2

I opted for keeping the defaults because 1) it'll be much easier to test all the backends by simply changing one `REACT_APP_SOURCE` value; 2) I think these details are kinda similar to the details we have for Site Search and App Search that lead to swiftype.com; 3) worst-case scenario - hackers will get access to a mostly empty deployment with some non-sensitive data.

examples/sandbox/src/App.js
yakhinvadim: Yeah, I was agonizing over this one. I chose to go this way to simplify the dev experience: it's much easier to import _everything_ from the Search UI state rather than figuring out what part of data should come from the connector and what part - from the Search UI state. 

But that's not a strong opinion. I can see how we're combining two different things in one object. 

Maybe putting the connector state under a separate field would resolve your concern? Like:

```
mapContextToProps={({ wasSearched, connectorState }) => ({
 
...
...
...
 
!connectorState.isLoggedIn && (
  <EuiModal
    onClose={() => {}}
    maxWidth={354}
    className="login-modal"
  >
```

Let me know if you still prefer to not put it in the Search UI state.

joemcelroy: Jason's idea is pretty clean as you could provide a hook that customers export

```javascript
  import { useAuthenticationState } from '@elastic/react-search-ui'

   const { isLoggedIn } = useAuthenticationState()
```

I think for now keeping the state within the Search UI is ok but agree there is likely a better route for this. Personally I would vote to keep as is (due to low code impact right now) and then think more in depth how authentication is handled for customers who require authentication in general (I see authentication stretching to not only workplace problems but also app search use-cases where requests require some sort of auth)

packages/search-ui-workplace-search-connector/bin/build-js
```diff
@@ -0,0 +1,7 @@
+# clear directory
+rm -rf ./lib
```
joemcelroy, line 2: can remove this line

packages/search-ui-workplace-search-connector/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "bugs": {
+    "url": "https://github.com/elastic/search-ui/issues"
+  },
+  "devDependencies": {
+    "npm-run-all": "^4.1.5",
+    "rimraf": "^2.6.3",
+    "typescript": "^4.5.4"
+  },
+  "dependencies": {
+    "@elastic/app-search-javascript": "^7.13.1",
```
joemcelroy, line 41: Should be able to remove dep

packages/search-ui-workplace-search-connector/src/__tests__/WorkplaceSearchAPIConnector.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  AutocompleteQuery
+} from "@elastic/search-ui";
+import { DEFAULT_STATE } from "@elastic/search-ui";
+import exampleResponse from "./exampleResponse.json";
+import { adaptResponse } from "../responseAdapter";
+
+jest.mock("../responseAdapter");
+
+(window as any).fetch = jest.fn();
+(window as any).fetch.mockReturnValue({
```
joemcelroy, line 19: Not a biggie right now but I prefer using something like nock to still allow real HTTP calls to be performed but allow us to control the responses cleaner.

packages/search-ui/src/SearchDriver.ts
```diff
@@ -455,6 +459,17 @@ class SearchDriver {
           }
         },
         (error) => {
+          if (error.message === INVALID_CREDENTIALS) {
```
joemcelroy, line 462: Assuming that an expired token puts the search state into an error state, one thing we should consider is the experience this has to our customers. In my opinion, I would say an invalid token should continue to show previous search results & show login modal overlaid. Treating it like an explicit error might show the login + error state.

examples/sandbox/src/App.js
JasonStoltz: I'll defer to ya'll on this one. If it eases developer experience then it's not a bad idea I suppose. We could change this API in a future breaking change.

We'll need to document it as new state that will only be populated depending on the connector that you use.

JasonStoltz:
@yakhinvadim In your epic for this connector, we should add "Disjunctive facets" as a task. We should have that for this connector, we'll need it for the starter.

examples/sandbox/src/App.js
yakhinvadim: Sounds good. I'll add a note to the issue to revisit this.

packages/search-ui-workplace-search-connector/src/WorkplaceSearchAPIConnector.ts
```diff
@@ -0,0 +1,338 @@
...
+  constructor({
+    kibanaBase,
+    enterpriseSearchBase,
+    redirectUri,
+    clientId,
+    beforeSearchCall = (queryOptions, next) => next(queryOptions),
+    beforeAutocompleteResultsCall = (queryOptions, next) => next(queryOptions),
+    beforeAutocompleteSuggestionsCall = (queryOptions, next) =>
+      next(queryOptions),
+    ...rest
```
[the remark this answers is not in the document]
yakhinvadim: 👍 Removed https://github.com/elastic/search-ui/pull/650/commits/e0ffe724a06d995c39372271bbfa68434827dec4

packages/search-ui-workplace-search-connector/package.json
yakhinvadim: Nice catch, removed https://github.com/elastic/search-ui/pull/650/commits/12a480cfd65e2563d9f8972176d785b7869bc887

packages/search-ui-workplace-search-connector/bin/build-js
yakhinvadim: 👍 Removed https://github.com/elastic/search-ui/pull/650/commits/e39a83a0e22911acdc8154845ead1038d2b3549e

packages/search-ui-workplace-search-connector/src/__tests__/WorkplaceSearchAPIConnector.test.ts
yakhinvadim: Yeah, the tests I added are not very good. Happy to revisit later.

packages/search-ui/src/SearchDriver.ts
yakhinvadim: I agree. That's the way it works now.
Screen Cast 2022-03-03 at 9 51 03 AM

Maybe we could not make it an Error in the code as well.

joemcelroy (approved):

PR #659, opened 2022-03-14
Divide connector examples into Pages

joemcelroy:
Following Tasks Done:
- clean up the create react app (remove service worker, boilerplate test spec, static assets)
- Create pages per connector example. Now 3 distinct pages: app-search, workplace, elasticsearch (kept app / site search in one page with the pre-existing env variable to switch modes)
- inlined the configuration into each page
- moved the workplace specific css into the workplace specific page folder
- Added a redirect from / -> /app-site-search to keep existing default behaviour of app-search being displayed.


yakhinvadim (approved):
🙌

PR #664, opened 2022-03-15
Add Workplace Search analytics endpoints

yakhinvadim:
This PR is mostly a clone of the original Joe's PR: https://github.com/yakhinvadim/search-ui/pull/2
The last commit adds all the required params to the API and updates the Autocomplete method which didn't work before.

Closes https://github.com/elastic/enterprise-search-team/issues/1482


packages/search-ui-workplace-search-connector/src/WorkplaceSearchAPIConnector.ts
```diff
@@ -224,33 +274,17 @@ class WorkplaceSearchAPIConnector {
...
-      // const response = await this.client.search(query, newOptions);
+      const responseJson = await this.performFetchRequest(apiUrl, {
+        query,
+        ...newOptions
+      });
       return adaptResponse(
         responseJson,
         buildResponseAdapterOptions(queryConfig)
       );
+      // const response = await this.client.search(query, newOptions);
```
joemcelroy, line 287: I think we should remove these uncommented lines as likely dont add value keeping them here.

yakhinvadim: Yeah, nice catch, missed it.

yakhinvadim:
Assuming you don't have additional comments 🙂 
Let me know if there's anything else I could fix in the follow-up PRs.

PR #665, opened 2022-03-15
Workplace Search header and logout

yakhinvadim:
This PR adds a header to the Workplace Search sandbox and implements a logout button.

Clicking on the logout button does two things:
1) Removes the token from the localstorage
2) Redirects user to [ent-search]/logout which invalidates the ent-search session and redirects to [kibana]/logout which invalidates the kibana session

The logout method was implemented on the Workplace Search connector and passed as an action to search-ui core actions.

https://user-images.githubusercontent.com/11838280/158480315-7cb9d316-6bcb-41dc-af72-95cef9227a33.mp4

### DIFF
--- a/packages/search-ui-workplace-search-connector/src/WorkplaceSearchAPIConnector.ts
+++ b/packages/search-ui-workplace-search-connector/src/WorkplaceSearchAPIConnector.ts
@@ -111,6 +111,9 @@ class WorkplaceSearchAPIConnector {
     authorizeUrl: string;
     isLoggedIn: boolean;
   };
+  actions: {
+    logout: () => void;
+  };
 
   /**
    * @param {Options} options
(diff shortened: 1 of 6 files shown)


packages/search-ui/src/SearchDriver.ts
```diff
@@ -163,6 +163,12 @@ class SearchDriver {
       },
       {}
     ) as actions.SearchDriverActions;
+
+    this.actions = {
```
joemcelroy, line 167: I feel this might need more thought. With this change, we are allowing some connectors to support more functionality than others which can be a little gotcha here with some components depending on particular connectors being used. I also think `logout` would not exist within the types here as its not part of the SearchDriverAction types.

packages/search-ui-workplace-search-connector/src/WorkplaceSearchAPIConnector.ts
```diff
@@ -157,6 +166,14 @@ class WorkplaceSearchAPIConnector {
       isLoggedIn: !!this.accessToken || false
     };
 
+    this.actions = {
+      logout: () => {
+        this.accessToken = null;
+        this.state.isLoggedIn = false;
```
joemcelroy, line 172: i've forgotten now - how does the connector state sync with SearchDriver state and trigger an view update?

joemcelroy (approved):
Approving for now as we plan to address these issues in a future release :)

PR #666, opened 2022-03-16
Elasticsearch-connector: Facet Value Sort Feature 

joemcelroy:
Supports the ability to sort facet values alphabetically. Configuration is consistent with the facet field configuration API. Also involved building the feature into Searchkit https://github.com/searchkit/searchkit/pull/1092.

Also fixed an annoying issue where the search placeholder for Facet is now using the label value, not the field. Was putting in "states.keyword" in the placeholder and really annoyed me 😂 

```javascript
facets: {
      "world_heritage_site.keyword": { type: "value", sort: "count" },
      "states.keyword": { type: "value", size: 30, sort: "value" } // using the sort attribute
}
``` 

image

### DIFF
--- a/packages/search-ui-elasticsearch-connector/src/handlers/search/Configuration.ts
+++ b/packages/search-ui-elasticsearch-connector/src/handlers/search/Configuration.ts
@@ -63,7 +63,8 @@ function buildConfiguration(
             field: facetKey,
             label: facetKey,
             size: facetConfiguration.size || 20,
-            multipleSelect: isDisJunctive
+            multipleSelect: isDisJunctive,
+            order: facetConfiguration.sort || "count"
           })
         );
       } else if (
--- a/examples/sandbox/src/pages/elasticsearch/index.js
+++ b/examples/sandbox/src/pages/elasticsearch/index.js
@@ -66,7 +66,7 @@ const config = {
     disjunctiveFacets: ["acres", "states.keyword", "date_established", "location"],
     facets: {
       "world_heritage_site.keyword": { type: "value" },
-      "states.keyword": { type: "value", size: 30 },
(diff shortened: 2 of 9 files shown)


packages/search-ui-elasticsearch-connector/src/handlers/search/Configuration.ts
```diff
@@ -63,7 +63,8 @@ function buildConfiguration(
             field: facetKey,
             label: facetKey,
             size: facetConfiguration.size || 20,
-            multipleSelect: isDisJunctive
+            multipleSelect: isDisJunctive,
+            order: facetConfiguration.sort || "count"
```
yakhinvadim, line 67: Just wondering, what do you think about defaulting to alphabetical?

yakhinvadim:
> Also fixed an annoying issue where the search placeholder for Facet is now using the label value, not the field. Was putting in "states.keyword" in the placeholder and really annoyed me 😂

Just wanted to check that this isn't a breaking change: will it change the displayed placeholder value for the existing Search UI projects after the upgrade?

yakhinvadim (approved):
LGTM pending the breaking change question. Feel free to merge if it's not a breaking change.

packages/search-ui-elasticsearch-connector/src/handlers/search/Configuration.ts
joemcelroy: yeh I went with count as thats the default behaviour when not specified in app-search / swifttype.

joemcelroy:
i dont see the facet change as a breaking change as its more of a bug - we shouldn't of used a field name for a placeholder label. so the label currently Label is a required field so shouldn't need the developer to make changes and therefore I see it as a minor fix.

PR #696, opened 2022-04-04
Make state property on APIConnector interface optional

yakhinvadim:
It is only needed in the Workplace Search connector and it's possible that it will be removed from that interface at some point.

This is a follow-up to https://github.com/elastic/search-ui/pull/682, where the state property was added to APIConnector interface.


elasticdocs:

* [View deploy preview](https://search-i93spx8ie-elastic-dev.vercel.app)

* Built with commit 0430d5cacba3351f1b7cd37153e06f3bfb95d973

joemcelroy (approved):

PR #713, opened 2022-04-20
Move from local to Elastic doc system

goodroot:
Removes docsmobile library, only content remains.


elasticdocs:

* [View deploy preview](https://search-lgwk5y6o8-elastic-dev.vercel.app)

* Built with commit 18253af2b784fa0de39bdf5ffa41ae794fb32dbd

goodroot:
@JasonStoltz This is ready to merge.

FYI, this is might get a bit messy.

Next steps are:

1. Merge this
2. Setup the pipeline for Elastic docs 
3. Test
4. Provide link to new home @ docs-staging.elastic.co/search-ui
5. Confirm we're happy with flow, "launch" to docs.elastic.co/search-ui

gtback:
@goodroot do we want to use `docs.elastic.co/en/search-ui` (note the **/en/**) for consistency with the Integrations docs, even if we never plan to translate these docs?

goodroot:
@gtback Vercel can apply rewrites to take care of this prefix. 

It's a slightly different model that doesn't "enforce" the language prefix.

For example: `example.com/en/page` is the same as `example.com/page`. Landing on the page with the `/en/` specified does a non-redirect rewrite of the page contents to `/page`. In the event we add languages, it's expanded then through rewrites vs. redirects, or firm URL structures.

joemcelroy (approved):

PR #718, opened 2022-04-25
[Client Telemetry] Enables reporting meta headers for client usage based telemetry

joemcelroy:
This PR adds to the request a http header `x-elastic-client-meta` which contains information about the client thats used to perform the request on cloud. This header is used to track client usage on cloud. The value follows a spec [documented here](https://github.com/elastic/clients-team/blob/master/knowledgebase/specifications/client-meta-header.md) and is validated based on a regex mentioned in the spec doc. 

`-(ws,es)-connector` is included to differentiate it between which connector is being used. If the client is run within the browser, we use `browser` for the JS version value. If its running in node context, use the nodejs version.  For workplace connector, we hardcode `browser` as the connector was never designed to run within nodejs.


elasticdocs:

* [View deploy preview](https://search-o2imtk68r-elastic-dev.vercel.app)

* Built with commit e5a9ec120dd8892829c04b59111ba9c8677985f7

packages/search-ui-workplace-search-connector/src/__tests__/WorkplaceSearchAPIConnector.test.ts
```diff
@@ -331,6 +334,55 @@ describe("WorkplaceSearchAPIConnector", () => {
...
+        filters: [
+          {
+            field: "date_made",
+            values: ["yesterday"],
+            type: "all" as const
+          }
+        ]
+      };
+
+      await subject(state, queryConfig);
```
yakhinvadim, line 373: Maybe we could replace this with just `await subject();`? It looks like `state` and `queryConfig` values are not important for this test case.

packages/search-ui-workplace-search-connector/src/__tests__/WorkplaceSearchAPIConnector.test.ts
```diff
@@ -331,6 +334,55 @@ describe("WorkplaceSearchAPIConnector", () => {
...
+
+      expect(global.fetch).toHaveBeenCalledTimes(1);
+      const metaHeader = getLastFetchCall().headers["x-elastic-client-meta"];
+      expect(metaHeader).toEqual(
+        "ent=1.11.0-ws-connector,js=browser,t=1.11.0-ws-connector,ft=universal"
```
yakhinvadim, line 378: I see that a similar test for onAutocomplete uses LIB_VERSION instead of hardcoded values. Maybe we could use it here as well to avoid updating tests in each version?

yakhinvadim (changes requested):
LGTM overall, just a couple nits:

packages/search-ui-workplace-search-connector/src/__tests__/WorkplaceSearchAPIConnector.test.ts
joemcelroy: thanks for spot! this should use `LIB_VERSION`

yakhinvadim (approved):
🎉

joemcelroy:
tested and verified the request headers are being sent on both app-search (now with app-search-javascript updates) and connectors (tested the ES one only)
Screenshot 2022-04-25 at 16 11 49
Screenshot 2022-04-25 at 16 20 34

joemcelroy:
tested with workplace-connector too. Works as expected. 

Waiting for these telemetry events to reach the telemetry dashboard. Once confirmed, will merge

PR #737, opened 2022-05-02
Improve linting

yakhinvadim:
This PR updates several things related to linting and prettier: mostly improved configs, commands and fixes errors. All the changes are described in commit messages.

This is not the end-state, there's still room for improvement, but this felt like a good place to pause. 

Once this is merged, I'll work on the second PR that will include: running prettier on the entire codebase and fixing all eslint errors, improving pre-commit hooks to utilize the new commands.


packages/react-search-ui/.gitignore
```diff
@@ -1,2 +0,0 @@
-lib
-es
```
yakhinvadim, line 2: Missed in the first commit

joemcelroy (approved):

PR #740, opened 2022-05-03
Improve linting 2

yakhinvadim:
This PR is a follow-up for https://github.com/elastic/search-ui/pull/737.

It mainly does two things:
- Sets up husky and lint-staged to run `lint --fix` and `prettier --write` on every commit. 
- Runs prettier and eslint on the entire codebase and fixes all warnings.

Commit descriptions have more detailed diff explanations.

---

The docs build is failing with:
```
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^15.3.0 || ^16.0.0" from react-select@2.4.4
```
I haven't looked yet into solving this. I think I'd prefer to address this separately.


.ci/jobs/search-ui.yml
```diff
@@ -13,10 +13,9 @@
           repo: search-ui
           repo-owner: elastic
           branch-discovery: 'no-pr'
-          discover-pr-origin: current
+          discover-pr-origin: 'both'
           discover-pr-forks-strategy: 'merge-current'
           discover-pr-forks-trust: 'permission'
-          discover-pr-origin: 'merge-current'
```
yakhinvadim, line 19: Duplicate keys were causing prettier to error out. 

I updated the config with the option that I thought made the most sense:
https://jenkins-job-builder.readthedocs.io/en/latest/project_githuborg.html

> discover-pr-origin (str) – Discovers pull requests where the origin repository is the same as the target repository. Valid options: merge-current, current, both, false. (default ‘merge-current’)

yarn.lock
```diff
@@ -57,7 +57,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.0.0", "@babel/core@^7.1.0", "@babel/core@^7.11.1", "@babel/core@^7.12.3", "@babel/core@^7.16.0", "@babel/core@^7.4.5", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
+"@babel/core@^7.1.0", "@babel/core@^7.11.1", "@babel/core@^7.12.3", "@babel/core@^7.16.0", "@babel/core@^7.4.5", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
```
yakhinvadim, line 60: I don't know why, but I got this diff after simply running `yarn`. I decided to commit it separately.

package.json
```diff
@@ -46,10 +35,8 @@
     "eslint-plugin-jest": "^26.0.0",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-react": "^7.28.0",
-    "husky": "^1.3.0",
     "jest": "^27.4.7",
     "lerna": "^4.0.0",
-    "lint-staged": "^9.2.0",
```
yakhinvadim, line 52: We were using really old versions, so reinstalling seemed simpler than migrating.

JasonStoltz (approved):

PR #754, opened 2022-05-16
[Elasticsearch-connector] post process query support

joemcelroy:
Enables functionality to mutate / read the elasticsearch query before its requested. Useful for advanced functionality such as overriding the query. 

```js
const connector = new ElasticSearchAPIConnector(
  {
    host: "https://example-host.es.us-central1.gcp.cloud.es.io:9243",
    index: "national-parks",
    apiKey: "exampleApiKey"
  },
  (esQuery) => {
    console.log("postProcess Query Call", esQuery); // logging out the esQuery before sending to Elasticsearch
    if (!esQuery.query) return esQuery;

    const bestMatchQuery = esQuery.query.bool.should[0].multi_match;
    esQuery.query = {
      multi_match: {
        query: bestMatchQuery.query,
        fields: bestMatchQuery.fields
      }
    };

    return esQuery;
  }
);
```

https://github.com/elastic/search-ui/pull/754


elasticdocs:
:rocket: Built _docs-elastic-co_ successfully!

* https://docs-elastic-mv3si8r2t-elastic-dev.vercel.app

* Built with commit bb3701d85f41fb23972655486231d87d69c9ba4a

examples/sandbox/src/pages/elasticsearch/index.js
```diff
@@ -24,15 +24,21 @@ import {
...
+      "https://search-ui-sandbox.es.us-central1.gcp.cloud.es.io:9243",
+    index: process.env.REACT_ELASTICSEARCH_INDEX || "national-parks",
+    apiKey:
+      process.env.REACT_ELASTICSEARCH_API_KEY ||
+      "SlUzdWE0QUJmN3VmYVF2Q0F6c0I6TklyWHFIZ3lTbHF6Yzc2eEtyeWFNdw=="
+  },
+  (esQuery) => {
+    console.log("postProcess Query Call", esQuery); // eslint-disable-line no-console
+    return esQuery;
+  }
```
yakhinvadim, line 40: Since this is more of an advanced use case, what do you think about removing this from the sandbox? I worry that it might be considered required by users and will be copied to implementations.

yakhinvadim (commented):
I plan to do another review, but wanted to give this feedback early.

I was a little confused between terms "query", "esQuery" and "request body" in docs and code. It feels like sometimes there were used interchangeably, and sometimes they meant different things. I think using the exact terms everywhere would be very helpful for understanding this functionality.

---

Oh, I would also try to avoid using abbreviation "ES", because it could be read as Elasticsearch and Enterprise Search.

examples/sandbox/src/pages/elasticsearch/index.js
joemcelroy: sure lets remove it

joemcelroy:
> I plan to do another review, but wanted to give this feedback early.
> 
> I was a little confused between terms "query", "esQuery" and "request body" in docs and code. It feels like sometimes there were used interchangeably, and sometimes they meant different things. I think using the exact terms everywhere would be very helpful for understanding this functionality.

Thank you! I've updated it to just use `requestBody` concepts.

> Oh, I would also try to avoid using abbreviation "ES", because it could be read as both Elasticsearch and Enterprise Search.

I will try, not the easiest with method / variable names becoming super long that they are unreadable if used in full.

docs/connectors-api-elasticsearch.mdx
```diff
@@ -208,3 +208,35 @@ const config = {
...
+## Customise the Elasticsearch Request Body
+
+Elasticsearch connector allows you to customise the Elasticsearch request body before its performed on Elasticsearch. This is useful if you want to customise the query or options before the request is sent to Elasticsearch.
+
+This is an advanced option, the underlying query may change between versions and reading from / mutating the query is brittle, so please be aware to use this sparingly and let us know what you want to achieve through github issues.
+
+Example below is overriding the `query` section of the Elasticsearch request body.
+
+```js
+const connector = new ElasticSearchAPIConnector(
```
yakhinvadim, line 221:
```suggestion
const connector = new ElasticsearchAPIConnector(
```

packages/search-ui-elasticsearch-connector/src/types.ts
```diff
@@ -0,0 +1,4 @@
+export type ESRequestBody = any;
+export type PostProcessRequestBodyFn = (
+  requestBody: ESRequestBody
+) => ESRequestBody;
```
yakhinvadim, line 4: I think we could differentiate Elasticsearch and Enterprise Search in variable names by calling the first `Es` (1 word) and second - `ES` (2 words)
```suggestion
export type EsRequestBody = any;
export type PostProcessRequestBodyFn = (
  requestBody: EsRequestBody
) => EsRequestBody;
```

docs/connectors-api-elasticsearch.mdx
```diff
@@ -208,3 +208,35 @@ const config = {
...
+  {
+    host: "https://example-host.es.us-central1.gcp.cloud.es.io:9243",
+    index: "national-parks",
+    apiKey: "exampleApiKey"
+  },
+  (requestBody) => {
+    console.log("postProcess Query Call", requestBody); // logging out the esQuery before sending to Elasticsearch
```
yakhinvadim, line 228:
```suggestion
    console.log("postProcess requestBody Call", requestBody); // logging out the requestBody before sending to Elasticsearch
```

yakhinvadim (commented):
Apologies for late review! LGTM overall, just a couple naming suggestions:

packages/search-ui-elasticsearch-connector/src/types.ts
joemcelroy: TBH im not super happy with "ES" being in the type name. Was thinking about having a seperate type file for ES and using same names as the official elasticsearch types https://github.com/elastic/elasticsearch-js/blob/main/src/api/types.ts#L937

yakhinvadim: Yeah, good idea.
Perhaps later we could even explore exporting types from that client.

yakhinvadim (approved):
LGTM 👍 
Just 1 question

PR #768, opened 2022-06-20
[elasticsearch-connector] Implement basefilters

joemcelroy:
Fixes #768

Observations:
- App Search when configured with global filters (for example states: california) and one of the global filters is a disjunctive facet, the "disjunctive facet call" incorrectly removes the global filter, which means the facet is incorrectly showing all possible options, including aggregations to bring back results which do not have the state.
- app search does allow adding filters that are not facets

Proposal
  - Changes to search-ui
    - No longer remove the filters from queryConfig so the connector can access the filters applied outside of the UI

  - Changes to app-search connector
    - Continue the merging of the `queryConfig.filters` + `requestState.filters` and mutate `requestState.filters`
    - Do not use `queryConfig.filters` in the `onSearch` event handler (as the filters are already present in the `requestState.filters`)

  - Changes to elasticsearch-connector
    - `queryConfig.filters` are transformed into lucene queries and placed into searchkit as a baseFilter
    - `requestState.filters` will be pruned where the filters exist in the `queryConfig.filters` (due to the mergedFilters behaviour in search-ui) and then are added as normal searchkit filters

- create a ticket to address the App search disjunctive + global filters issue


packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.ts
```diff
@@ -386,7 +386,17 @@ describe("AppSearchAPIConnector", () => {
             {
               all: [
                 {
-                  date_made: "yesterday"
+                  title: "Acadia"
```
joemcelroy, line 389: This change seems a little odd but *i believe* the correct behaviour here. The test case here has two filters in the request state and one filter in the queryConfig. Previously it was only providing the filter in the `queryConfig` to the API. Now its passing all three filters to the API. I believe it should be passing all three filters to the API.

packages/search-ui/src/__tests__/SearchDriver.test.ts
```diff
@@ -334,9 +334,9 @@ describe("searchQuery config", () => {
       ]);
     });
 
-    it("will remove filters parameter from queryConfig since its merged into state", () => {
+    it("will not remove filters parameter from queryConfig", () => {
```
joemcelroy, line 337: this test was changed as the queryConfig.filters still remain as its needed by the connectors

packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.ts
joemcelroy: Thanks for the spot! So the testcase is incorrect here, the `date_made: "yesterday"` should of been present in `requestState.filters` too, as searchDriver runs `mergeFilter` which mutates the `requestState.filters` with filters from the `queryConfig.filters`

yakhinvadim (approved):
Working as expected now! Great work, Joe! 👍

PR #770, opened 2022-06-20
Fix broken watch command

yakhinvadim:
Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "start": "yarn --cwd examples/sandbox start",
     "test": "lerna run test --stream --no-private",
     "test-ci": "lerna run test-ci --stream --no-private",
-    "watch": "lerna run watch --parallel",
+    "watch": "lerna exec --no-private -- yarn run watch",
     "build": "lerna run build --no-private",
     "prettier-check": "prettier --check --ignore-unknown --ignore-path .gitignore .",
     "prettier-format": "prettier --write --ignore-unknown --ignore-path .gitignore .",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test": "lerna run test --stream --no-private",
     "test-ci": "lerna run test-ci --stream --no-private",
     "watch": "lerna run watch --parallel",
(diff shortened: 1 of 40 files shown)


package.json
```diff
@@ -10,8 +10,8 @@
     "start": "yarn --cwd examples/sandbox start",
     "test": "lerna run test --stream --no-private",
     "test-ci": "lerna run test-ci --stream --no-private",
-    "watch": "lerna run watch --parallel",
-    "build": "lerna run build --stream --scope @elastic/search-ui && lerna run build --stream --scope '{@elastic/react-search-ui*,@elastic/search-ui-*}'",
+    "watch": "lerna exec --no-private -- yarn run watch",
```
joemcelroy, line 13: tried this with `run` and didn't work, realised you likely hit this too! I then learnt about differences between exec and run :joy: 

its surprising it includes the root `package.json` in the execution. If we do introduce any private packages, we can change it to an `--exclude root` flag.

packages/react-search-ui-views/package.json
```diff
@@ -30,7 +31,7 @@
     "watch-js": "./bin/watch-js",
     "build-js": "./bin/build-js",
     "build": "run-s clean build-css build-js",
-    "watch": "npm run clean && run-p watch-css watch-js",
+    "watch": "npm run build && run-p watch-css watch-js",
```
joemcelroy, line 34: do you think we need a `run build` step before the `watch`? watch-js runs the same command as the build step, just includes a watch flag?

joemcelroy (approved):
LGTM, just some comments for my understanding :)

packages/react-search-ui-views/package.json
yakhinvadim: Exactly! That's what I tried at first but it didn't work for me. TBH I don't really follow what happens under the hood here, and just want to make this command work even at a cost of little inefficiency. 🤷

package.json
yakhinvadim: How do you know that it includes root `package.json` in the execution? My terminal gets cleared quickly and I can't check the logs :D

---

BTW, we do have private packages! examples/sandbox is considered private package because it's listed under packages: https://github.com/elastic/search-ui/blob/master/lerna.json#L2

Do you think we still need it there?

Issue #769, opened 2022-06-20
Upgrade react-input-autosize


spalger:
The latest version of `@elastic/react-search-ui` supports React 17, but the `react-input-autosize` dependency wasn't updated. v3.0.0 is the latest and it's [only real change](https://github.com/JedWatson/react-input-autosize/blob/master/HISTORY.md#v300--2020-12-11) is to support React 17. Could we get an update to remove one more version warning in the Kibana repo?

yakhinvadim:
Thanks for the report, Spencer. I'll bring this up with the team tomorrow.

yakhinvadim:
@spalger How can I reproduce the warning? I tried `yarn kbn bootstrap` but didn't see it there.

spalger:
I had to trigger the warning by upgrading to React 17 in a branch

yakhinvadim:
We plan to release this early next week and then upgrade the dependency in Kibana. I'll post the update here.

yakhinvadim:
[The upgrade](https://github.com/elastic/kibana/pull/135363) didn't go as smoothly as I hoped, we need to dig into TypeScript errors. Ok if we do it next week?

spalger:
Sure thing, thank you

spalger:
Whoops!

yakhinvadim:
@spalger This is done! We updated Search UI in Kibana, it shouldn't import react-input-autosize anymore.

spalger:
Awesome, thank you!

PR #776, opened 2022-06-23
Upgrade react-select to remove warnings about unsafe react methods

yakhinvadim:
Closes https://github.com/elastic/search-ui/issues/769

Tested on all our examples, upgrade brings no visible changes.


packages/react-search-ui-views/src/__tests__/__snapshots__/ResultsPerPage.test.tsx.snap
```diff
@@ -9,17 +9,14 @@ exports[`renders correctly when there is a selected value 1`] = `
...
     className="sui-select sui-select--inline"
     classNamePrefix="sui-select"
     components={
       Object {
         "Option": [Function],
       }
     }
-    defaultInputValue=""
-    defaultMenuIsOpen={false}
-    defaultValue={null}
```
yakhinvadim, line 22: We're not using these props and do not allow Search UI users to target them, so I think the diff is safe.

joemcelroy:
LGTM. worth mentioning in our release notes the upgrade guide for react-select https://react-select.com/upgrade#upgrade-guide. We can release this on its own.

joemcelroy (approved):

Issue #782, opened 2022-06-28
[Elasticsearch-Connector] Access inner_hits


kokolem:
**Is your feature request related to a problem? Please describe.**
When using a nested query:

```js
const connector = new ElasticsearchAPIConnector({
  host: "http://localhost:9200",
  index: "index"
}, (requestBody, requestState) => {
  if (!requestState.searchTerm) return requestBody;

  requestBody.query = {
    nested: {
      path: "nested_object",
      query: {
        multi_match: {
          query: requestState.searchTerm,
          fields: ["field"]
        }
      },
      inner_hits: {}
    }
  }

  return requestBody;
})
```

there is currently no way to access `inner_hits` of the results.

**Describe the solution you'd like**
It would be nice to be able to access inner hits from the `result` parameter provided to `resultView`.

**Describe alternatives you've considered**
Unfortunately I haven't found any way to access inner_hits. Suggestions are welcome.

joemcelroy:
Thanks @kokolem  - an interesting use-case. I dont have a solution or workaround right now. Likely the best solution is providing the raw elasticsearch item hit with the result object so that you can render the nested documents.

I think long term we want to make nested document support 1st class. Facets and query to support nested objects.

joemcelroy:
released in 1.17.0

PR #785, opened 2022-06-30
Provide raw ES hit in result

kokolem:
## Description
Providing the raw elasticsearch item hit with the result object as part of the `_meta` attribute.

## List of changes

- [packages/search-ui-elasticsearch-connector/src/handlers/common.ts](https://github.com/elastic/search-ui/compare/master...kokolem:elastic-search-ui:master#diff-71108dd9ee9c2f539db42708446215b2674935023e82248b1907674d1b06584f)

include `item.rawHit` as part of `_meta`

- [packages/search-ui-elasticsearch-connector/src/handlers/autocomplete/\_\_tests\_\_/index.test.ts](https://github.com/elastic/search-ui/pull/785/commits/14e39ed2e49b6780b12cda486caf2f7b34f48485#diff-f920b738d9c29a8242cad31ef4dca1e643e7ffeff66ed34c75d156175d5df774)
- [packages/search-ui-elasticsearch-connector/src/handlers/search/\_\_tests\_\_/Response.test.ts](https://github.com/elastic/search-ui/pull/785/commits/14e39ed2e49b6780b12cda486caf2f7b34f48485#diff-aa00e202fde8696b009ccc898050fe85c5fc04fd225f1d9109c0a945eedd5b38)
- [packages/search-ui-elasticsearch-connector/src/handlers/search/\_\_tests\_\_/index.test.ts](https://github.com/elastic/search-ui/pull/785/commits/14e39ed2e49b6780b12cda486caf2f7b34f48485#diff-1f62913a7d8b75e759bae11a439847a39f26794367558e634dc4647ac2a7a65b)

update tests to expect the new `raw` attribute as part of `_meta`

## Associated Github Issues
https://github.com/elastic/search-ui/issues/782


packages/search-ui-elasticsearch-connector/src/handlers/common.ts
```diff
@@ -22,7 +22,8 @@ export function fieldResponseMapper(
     {
       id: { raw: item.id },
       _meta: {
-        id: item.rawHit._id
+        id: item.rawHit._id,
+        raw: item.rawHit
```
kokolem, line 26: This causes `isFieldValueWrapper` to return `true` for `_meta`, 
https://github.com/elastic/search-ui/blob/17488c45f21a496ad5cfbe1e09e624240647b93b/packages/react-search-ui-views/src/Result.tsx#L32-L38
which means we get `"_meta": [object Object]` in the results.

kokolem: This can be solved by adding a type check:
```typescript
function isFieldValueWrapper(object) {
  return (
    object &&
    (Object.prototype.hasOwnProperty.call(object, "raw") && typeof object["raw"] != "object" ||
      Object.prototype.hasOwnProperty.call(object, "snippet"))
  );
}
```

or simply renaming the property to something like `raw_es_hit`. What do you think is the best solution?
Also, it might be worth to write a test for this.

yakhinvadim:
@joemcelroy Just a quick ping in case this PR got lost in notifications 🙂

packages/search-ui-elasticsearch-connector/src/handlers/common.ts
kokolem: Done.

joemcelroy (approved):

PR #794, opened 2022-07-11
777 fix filters

### LINKED ISSUE #777: [Elasticsearch-connector] setFilter not working
**Describe the bug**
The setFilter action is not building the right Elasticsearch queries.

**To Reproduce**

Trigger setFilter for a long field

```
    setFilter("precio", {
      name: "precio",
      from: rangePrices[0],
      to: rangePrices[1],
    });
```

**Expected behavior**
A "range" query should trigger because the filter value is an object, but a term query is triggered instead. Also, the term query throws an error because "term" does not support "name" which is required in setFilter.

**Screenshots**

Backend: Elasticsearch
Packages: search-ui-elasticsearchsearch-connector,

joemcelroy:
fixes #777

### DIFF
--- a/packages/search-ui-elasticsearch-connector/package.json
+++ b/packages/search-ui-elasticsearch-connector/package.json
@@ -41,6 +41,6 @@
   "dependencies": {
     "@elastic/elasticsearch": "^8.2.1",
     "@elastic/search-ui": "1.15.0",
-    "@searchkit/sdk": "^3.0.0-canary.57"
+    "@searchkit/sdk": "^3.0.0-canary.58"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ storybook-static
 package-lock.json
 packages/**/src/version.ts
 packages/**/lib/*
+coverage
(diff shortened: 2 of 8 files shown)


packages/search-ui-elasticsearch-connector/package.json
```diff
@@ -41,6 +41,6 @@
   "dependencies": {
     "@elastic/elasticsearch": "^8.2.1",
     "@elastic/search-ui": "1.15.0",
-    "@searchkit/sdk": "^3.0.0-canary.57"
+    "@searchkit/sdk": "^3.0.0-canary.58"
```
JasonStoltz, line 44: Why do you use the canary versions in SearchKit?

JasonStoltz (commented):
I see:
1) Code
2) Tests
3) Updated example

All looks well to me. Is there anything we need to note in our docs about this?

packages/search-ui-elasticsearch-connector/package.json
joemcelroy: searchkit v3 not yet pass pre-release yet so still on canaries right now.

JasonStoltz (dismissed):
Hey Joe, I thought I added my review here but it looks like I did not. Everything looks good I'm just wondering if you need to update anything in docs?

joemcelroy:
> Hey Joe, I thought I added my review here but it looks like I did not. Everything looks good I'm just wondering if you need to update anything in docs?

Yeh its a good shout. I've added it to the connector api docs + another known issue. Hope these changes are ok now.

JasonStoltz (approved):

PR #805, opened 2022-07-18
Add Vue.js example

yakhinvadim:
This PR:
- Moves Vue.js demo from a separate repo to this repo
- Updates demo from Vue 2 to Vue 3
- Updates demo to use Cloud data
- Adds a guide to the docs

image

https://user-images.githubusercontent.com/11838280/179609176-f6d7a355-f31d-4014-aca5-be1f00f6666b.mp4

The biggest part of the PR (150000 LOC) is 2 jsons with data. Having the data in the repo will allow people to quickly ingest it into their cloud deployments. Most of the remaining diff is copy-paste from original repo https://github.com/elastic/vue-search-ui-demo

The demo in docs is currently unavailable since it points to master branch and this PR isn't merged yet.
But you can check it out separately: https://codesandbox.io/s/github/yakhinvadim/search-ui/tree/vue-example-1/examples/vue?file=/src/components/SearchSection.vue


rejwan-ahamed:
Can you provide the API link, please.

package.json
```diff
@@ -8,6 +8,8 @@
   ],
   "scripts": {
     "start": "yarn --cwd examples/sandbox start",
+    "start-vue": "yarn --cwd examples/vue dev",
```
joemcelroy, line 11: shall we remove this task?

package.json
```diff
@@ -8,6 +8,8 @@
...
+    "start-docs": "cd ../docs.elastic.co/ && yarn init-docs && yarn dev",
```
joemcelroy, line 12: same suggestion as above: dont think it should live in the react-search-ui-views package. Likely better on the root package or question whether it needs to exist.

examples/vue/cleanseCards.js
```diff
@@ -0,0 +1,30 @@
+const fs = require("fs");
```
joemcelroy, line 1: Suggestion: remove this file + src/data/cards-raw.json. rename `cards-cleansed.json` to `cards.json`. IMO if they want to ingest, they just need the parsed data documents.

package.json
yakhinvadim: It's already in the root package.json. 🙂 
I think it's convenient to be able to run all commands from the root of the repo rather than having to `cd` to specific directories.

examples/vue/cleanseCards.js
yakhinvadim: This data is constantly changing — Hearthstone keeps getting new cards, and some are getting removed. It doesn't _really_ matter for the demo, but I think if we update it once a year or so, it'll help to keep the demo fresher.

examples/vue/.gitignore
```diff
@@ -0,0 +1,28 @@
+# Logs
```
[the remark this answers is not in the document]
yakhinvadim: I decided to keep this and a couple of other files since they were added by the Vue project generator. They will help our users if they decide to copy the entire folder and use it as a starter project.

yakhinvadim:
@rejwan-ahamed 
> Can you provide the API link, please.

Sure, I added it to the README file here: https://github.com/elastic/search-ui/tree/ea35fa211c833fabb1ec09b52475e8f286a96f58/examples/vue/src/data

package.json
joemcelroy: Oops! I think I read the next change as the location.

joemcelroy (approved):

PR #807, opened 2022-07-19
Vue fixes

yakhinvadim:
A couple of minor fixes after the Vue demo merge, all are listed in commits.

### DIFF
--- a/examples/vue/src/App.vue
+++ b/examples/vue/src/App.vue
@@ -16,7 +16,7 @@ export default {
 </script>
 
 <style>
-@import "../node_modules/@elastic/react-search-ui-views/lib/styles/styles.css";
+@import "@elastic/react-search-ui-views/lib/styles/styles.css";
 
 body {
   margin: 0;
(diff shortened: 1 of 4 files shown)


examples/vue/src/App.vue
```diff
@@ -16,7 +16,7 @@ export default {
 </script>
 
 <style>
-@import "../node_modules/@elastic/react-search-ui-views/lib/styles/styles.css";
+@import "@elastic/react-search-ui-views/lib/styles/styles.css";
```
yakhinvadim, line 19: This import didn't work before when I was running the demo from the repo root, now both ways of running work.

yarn.lock
```diff
@@ -1681,34 +1681,6 @@
     uuid "^8.3.0"
     vfile "^4.2.0"
```
yakhinvadim, line 1683: The change below was actually added in the original Vue PR, so these lines were not here before. I guess they are not needed after all.
Screen Shot 2022-07-19 at 10 49 24 AM

joemcelroy (approved):

PR #859, opened 2022-09-26
Analytics Plugin Support

joemcelroy:
Enables Analytics support for Search UI.

- [x] Unit tests for code
- [x] Manually tested locally

### DIFF
--- a/packages/search-ui-analytics-plugin/README.md
+++ b/packages/search-ui-analytics-plugin/README.md
@@ -1,0 +1,3 @@
+# search-ui-analytics-plugin
+
+Part of the [Search UI](https://github.com/elastic/search-ui) project.
--- a/packages/search-ui-analytics-plugin/README.md
+++ b/packages/search-ui-analytics-plugin/README.md
@@ -1,3 +1,29 @@
 # search-ui-analytics-plugin
 
 Part of the [Search UI](https://github.com/elastic/search-ui) project.
+
+### Integration
+
+Enable Behaviorial analytics within kibana.
+
+Embed the analytics client onto the page. See the example integrations snippet thats provided to you once you created a collection
+
+```
+<script
+  src="<HOST>/analytics.js"
+  data-dsn="<HOST>/analytics/api/collections/test"
+  defer
+></script>
+```
+
+then integrate the plugin into the configuration
+
+```
+import AnalyticsPlugin from "@elastic/search-ui-analytics-plugin";
+
+const config = {
+  ...
+  plugins: [AnalyticsPlugin()],
+  ...
+}
+```
--- a/packages/search-ui-analytics-plugin/package.json
+++ b/packages/search-ui-analytics-plugin/package.json
@@ -1,0 +1,41 @@
+{
+  "name": "@elastic/search-ui-analytics-plugin",
+  "version": "1.17.1",
+  "description": "Behavorial Analytics middleware plugin",
+  "license": "Apache-2.0",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "lib/esm/index.d.ts",
+  "sideEffects": false,
+  "homepage": "https://docs.elastic.co/search-ui",
+  "directories": {
+    "lib": "lib",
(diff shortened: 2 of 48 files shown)


packages/search-ui-analytics-plugin/README.md
```diff
@@ -0,0 +1,29 @@
+# search-ui-analytics-plugin
```
JasonStoltz, line 1: We don't have to do this now, but perhaps in a future PR.

We should briefly explain what Behavioral Analytics is, and note that, for instance this is an experimental feature of Elastic that is not yet available.

Note that doing so will start sending events to the Elastic server, and maybe enumerate what those events are.

It would be also be super helpful to document somewhere in our docs something about "Analytics" and how you can use middleware to subscribe and handle all of these various events.

packages/search-ui-analytics-plugin/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@elastic/search-ui-analytics-plugin",
+  "version": "1.17.1",
```
JasonStoltz, line 3: I guess we don't really have the ability to release this as a pre-release or anything?

packages/search-ui-elasticsearch-connector/package.json
```diff
@@ -41,6 +41,6 @@
   "dependencies": {
     "@elastic/elasticsearch": "^8.2.1",
     "@elastic/search-ui": "1.17.1",
-    "@searchkit/sdk": "^3.0.0-canary.58"
+    "@searchkit/sdk": "^3.0.0"
```
JasonStoltz, line 44: Is there a functional consequence of this? If so, should we create a separate PR for it?

packages/search-ui-elasticsearch-connector/src/handlers/search/__tests__/Response.test.ts
```diff
@@ -43,8 +43,8 @@ describe("Search - Response", () => {
         label: "test",
         type: "value",
         entries: [
-          { label: "labeltest", count: "10" },
-          { label: "label2", count: "20" }
+          { label: "labeltest", count: 10 },
```
JasonStoltz, line 46: Also wondering if this should be in a separate PR.

packages/search-ui/src/__tests__/actions/trackAutocompleteSuggestionClickThrough.test.ts
```diff
@@ -0,0 +1,41 @@
+import { AutocompleteSuggestion } from "../..";
```
JasonStoltz, line 1: I think we need to document this in `api-core-actions.md`. Something we can do in a follow up PR.

JasonStoltz (approved):
I'm going to approve this so that I'm not holding you up, but please review my comments before merging.

Note that we could make the suggested documentation changes in a follow up PR.

packages/search-ui-analytics-plugin/package.json
joemcelroy: Yeh we dont have the ability yet for a pre-release without difficulty.

packages/search-ui/src/__tests__/actions/trackAutocompleteSuggestionClickThrough.test.ts
joemcelroy: Thanks for the spot. Will follow up tomorrow with documentation for that new action.

packages/search-ui-analytics-plugin/README.md
joemcelroy: Agreed. Definitely, documentation is a future task when analytics becomes publicly accessible. Il follow up with a PR to document it elsewhere for now.

packages/search-ui/src/__tests__/actions/trackAutocompleteSuggestionClickThrough.test.ts
joemcelroy: done

JasonStoltz (approved):

Issue #837, opened 2022-08-04
Queries are hitting 32 filter limit prematurely.


SaraMoad:
**Describe the bug**
- I am trying to use an array to filter which records should be returned.
- I am passing an array of ~500 values in. I am getting the error 'An unexpected error occurred: [400] Filters cannot have more than 32 filters'
- I am have done so trying to follow the guidance laid out [here](https://www.elastic.co/guide/en/app-search/current/filters.html#filters-arrays-vs-objects.): 

**To Reproduce**
` const config: SearchDriverOptions = {
    searchQuery: {
      facets: buildFacetConfigFromConfig(),
      disjunctiveFacets: ["level", "castingtime", "rarity", "ispfslegal", "traits", "subtype", "school", "area", "duration", "range", "savingthrow", "targets"],
      ...buildSearchOptionsFromConfig(),
      filters: [
        {
          field: "nethysid",
          type: "any",
          values: spellsToShowIDs
        }
      ]
    },
    autocompleteQuery: buildAutocompleteQueryConfig(),
    apiConnector: connector,
    alwaysSearchOnInitialLoad: true
  };`

**Expected behavior**
I expect the array spellsToShowIDs to filter successfully.

**Screenshots**
We can see the array size that i'm using, and the log of the config object here:
![image](https://user-images.githubusercontent.com/102942360/182867089-8326f23d-be56-42d1-9faa-bf8034dc2cc5.png)
We can see the request here:
![image](https://user-images.githubusercontent.com/102942360/182867032-f658df36-e683-485a-b9ca-f43151fe8e3c.png)

**Which backends and packages are you using:**
Backend: [App Search]
Packages: [react-search-ui, search-ui-app-search-connector]

joemcelroy:
thanks for the report! Will try get back to you, i suspect this is an issue with how we unroll filters together.  Linked with https://github.com/elastic/search-ui/issues/763

JamesLove002:
I’m also stuck with this issue!!

wentaoxu415:
HI @joemcelroy, is there any update on the fix for this issue?

joemcelroy:
no update yet @wentaoxu415. I will try get this prioritized.

joemcelroy:
I've started looking into this bug. Few things want to highlight:

- We are currently creating a new "filter" per value, even if it was grouped together inside values. This hits a ceiling fast with only 32 filters can be applied. 
[...]

PR #867, opened 2022-09-29
Efficent generation of filters for App Search Connector 

joemcelroy:
fixes https://github.com/elastic/search-ui/issues/837. 

- We are currently creating a new "filter" per value, even if it was grouped together inside values. This hits a ceiling fast with only 32 filters can be applied. 
- we should be using one filter per field with an **array** of values. https://www.elastic.co/guide/en/app-search/current/filters.html#filters-arrays-all-none
- ranges will not be able to be grouped together. Requires a filter per range type

### DIFF
--- a/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.ts
+++ b/packages/search-ui-app-search-connector/src/__tests__/AppSearchAPIConnector.test.ts
@@ -391,10 +391,7 @@ describe("AppSearchAPIConnector", () => {
             {
               all: [
                 {
-                  title: "Acadia"
-                },
-                {
-                  title: "Grand Canyon"
+                  title: ["Acadia", "Grand Canyon"]
                 }
               ]
             },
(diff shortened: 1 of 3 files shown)


joemcelroy:
@JasonStoltz This one you might know more about its history. Im still unsure why the rollup function split grouped filter values into individual filters. This PR should now increase the limit from 32 filter values o 1024 filter values and works on our sandbox environment.

JasonStoltz:
I think the reasoning is that these two are not necessarily functionally equivalent.


initial must match "additional values" AND "and values" AND "and even more values"
```
        all: [
          { initial: "additional values" },
          { initial: "and values" },
          { initial: "and even more values" }
        ]
```

initial must match "additional values" OR "and values" OR "and even more values"
```
        all: [
          {
            initial: ["additional values", "and values", "and even more values"]
          }
        ]
```

I think that as soon as you use the array syntax, the values implicitly become "OR"d. Let me do some quick testing to verify this.

JasonStoltz:
```
{
  "query": "",
  "filters": {
      "all": [{
          "states": ["Alaska", "California"]
      }]
  }
}
```

Is functionally equivalent to:
```
{
  "query": "",
  "filters": {
      "all": [{
          "any": [{
            "states": ["Alaska"]
          },{
            "states": ["California"]
          }]
      }]
  }
}
```

joemcelroy:
thanks @JasonStoltz that makes sense. I will update the PR to see if the filter is an "any" type and preserve the array of values, else create as before a filter for each value.

JasonStoltz (approved):

PR #868, opened 2022-10-04
fix the error while passing any data except string for normalization

### LINKED ISSUE #833: isFilterable={true} breaks when searching on a number
**Describe the bug**
When using a `<Facet isFilterable={true} />` on a field mapped to `NUMBER` results on.

 Uncaught TypeError: str.normalize is not a function

**To Reproduce**

1. Set `isFilterable =  true` on a `<Facet/>` 
2. Enter any value on the search field

**Expected behavior**
The list of values should be filtered to the value entered 

Packages:     
"@elastic/react-search-ui": "^1.16.0",
"@elastic/react-search-ui-views": "^1.16.0",
"@elastic/search-ui-app-search-connector": "^1.16.0"

**Culprit**
I believe that the problems resides on a helper that has the following code
```
export const accentFold = (str = "") => str.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
```

**Possible Solution**
We should either check if `typeof str === 'string` before running str.normalize("NFD") or try to cast to `string` if possible.

atiqueahmedziad:
## Description
When using a `<Facet isFilterable={true} />` on a field mapped to `NUMBER`, we get the following error:
`Uncaught TypeError: str.normalize is not a function`

## List of changes
1. If anything other than string passed as argument in `accentFold` function, it should return empty string.
2. Added a shield for facet values to be string.

## Associated Github Issues
Fixes #833

### DIFF
--- a/packages/react-search-ui/src/containers/Facet.tsx
+++ b/packages/react-search-ui/src/containers/Facet.tsx
@@ -93,9 +93,11 @@ export class FacetContainer extends Component<
 
     if (searchTerm.trim()) {
       facetValues = facetValues.filter((option) =>
-        accentFold(option.value)
-          .toLowerCase()
-          .includes(accentFold(searchTerm).toLowerCase())
+        typeof option.value === "string"
+          ? accentFold(option.value)
+              .toLowerCase()
+              .includes(accentFold(searchTerm).toLowerCase())
+          : false
       );
     }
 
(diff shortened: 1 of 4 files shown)


packages/react-search-ui/src/containers/Facet.tsx
```diff
@@ -93,9 +93,11 @@ export class FacetContainer extends Component<
 
     if (searchTerm.trim()) {
       facetValues = facetValues.filter((option) =>
-        accentFold(option.value)
-          .toLowerCase()
-          .includes(accentFold(searchTerm).toLowerCase())
+        typeof option.value === "string"
+          ? accentFold(option.value)
```
atiqueahmedziad, line 97: Found that `option.value` is not always string. Especially, when using a `<Facet isFilterable={true} />` on a field mapped to `NUMBER` results

atiqueahmedziad:
@JasonStoltz @joemcelroy review?

joemcelroy (approved):
Looks good. Thanks for your submission,

joemcelroy:
our ci pipeline seems to be broken for community contributions. Will investigate.

joemcelroy:
this was released under v1.18.1 🎉

PR #869, opened 2022-10-05
Routing options: Enables state URL customisation 

### LINKED ISSUE #534: fail to work with Next.js Client-side navigations ( next/router, next/link )
Hi. I'm using `search-ui` with my `Next.js` project. 
But I found that when I navigate away from the search page to another page using the `Client-side navigations`, I cannot go back by using the browser "back" button (I've only tried chrome.)

> It seems to be the only lib that support elastic-app-search. Is it correct?

How to replicate: 
1. Run this [demo repo]( https://github.com/jason-den/next-search-ui-demo)
2. visit search page `localhost:3001/search`
3. click `go home page` at the top of sidebar
4. when home page complete rendering, click your browser `back` button like this: 
image

Problematic result: The URL do change to `localhost:3001/search?......` but page remains in home page.
image

I'm not sure this issue come from `search-ui`, not from Next.js. I guess this issue has something relate to `Smart URL` -- which I want to understand more but couldn't find much documents. 

> BTW I also tried `search-ui` with `react router`, they worked well with each other. 

Any hint?

joemcelroy:
fixes #534 

Introduces a new configuration called `routingOptions`. This will allow you to override how state is serialised to url.

### DIFF
--- a/docs/guides-nextjs-integration.mdx
+++ b/docs/guides-nextjs-integration.mdx
@@ -1,10 +1,94 @@
 ---
 id: guides-nextjs-integration
 slug: /search-ui/guides/nextjs-integration
-title: NextJS with Elasticsearch Connector
-date: 2022-06-22
+title: NextJS Integration
+date: 2022-10-05
 tags: ["nextjs", "vercel"]
 ---
+
+### Integrating Search UI with NextJS Router
+
+Next JS is a popular React framework that provides a number of features out of the box, including server-side rendering, routing, and more.
+
+To take advantage of being able to save the search criteria in the URL, you need to override the existing routing options and use Next Router.
+
+Below is an example of how to do achieve this.
+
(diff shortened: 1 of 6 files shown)


docs/guides-nextjs-integration.mdx
```diff
@@ -119,3 +203,4 @@ Below is an example of this running within codesandbox.
   allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
   sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
 ></iframe>
+````
```
JasonStoltz, line 206: Is this line purposeful?

JasonStoltz (dismissed):

docs/guides-nextjs-integration.mdx
joemcelroy: thanks for spot!

PR #882, opened 2022-10-18
Fix facet filters for ranges

ckeboss:
## Description
Currently the ability to filter a Facet (`isFilterable` prop) does not work with range based facets for app-search, as the structure of the `value` property is different. Assuming this is intended, this change gives the ability to filter by that type of value.

If this is unintentional, we can change the connector to change the way the value structured in the response.

## List of changes
A bit of duck typing for the `value` property in `facetValues` to facilitate filtering better.

### DIFF
--- a/packages/react-search-ui/src/containers/Facet.tsx
+++ b/packages/react-search-ui/src/containers/Facet.tsx
@@ -92,13 +92,30 @@ export class FacetContainer extends Component<
     if (!facetValues.length && !selectedValues.length) return null;
 
     if (searchTerm.trim()) {
-      facetValues = facetValues.filter((option) =>
-        typeof option.value === "string"
-          ? accentFold(option.value)
-              .toLowerCase()
-              .includes(accentFold(searchTerm).toLowerCase())
-          : false
-      );
+      facetValues = facetValues.filter((option) => {
+        let valueToSearch;
+        switch (typeof option.value) {
+          case "string":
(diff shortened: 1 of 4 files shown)


packages/react-search-ui/src/containers/Facet.tsx
```diff
@@ -92,13 +92,30 @@ export class FacetContainer extends Component<
...
+        let valueToSearch;
+        switch (typeof option.value) {
+          case "string":
+            valueToSearch = accentFold(option.value).toLowerCase();
+            break;
+          case "number":
+            valueToSearch = option.value.toString();
+            break;
+          case "object":
+            valueToSearch =
```
joemcelroy, line 105: nitpick: could you use optional chaining to improve readability? 

`typeof option?.value?.name === "string"`

joemcelroy (commented):
Looks good other than small nitpick!

packages/react-search-ui/src/containers/Facet.tsx
ckeboss: Done

joemcelroy (approved):

joemcelroy:
thanks for your submission as always!

ckeboss:
@joemcelroy any idea when a new patch version will be published?

joemcelroy:
will try today or tomorrow. Unfortunately its a manual process right now.

PR #917, opened 2023-01-10
Nextjs Router Documentation

joemcelroy:
Updating the nextjs documentation to fix a couple of issues:
- if the search is present in a dynamic url, updating the url without the original params will fail. The fix is to extend the params with search ui params. Memo also needs to watch when router `isReady` so params can be provided.

In response to this discuss post: https://discuss.elastic.co/t/integrating-search-ui-with-nextjs-router-dynamic-routes/322248/5


elasticdocs:
:rocket: Built _docs-elastic-co_ successfully!

* https://docs-elastic-zsuu55n5x-elastic-dev.vercel.app

* Built with commit 47f454c08393aae1cdf98ff3d2e2f426d0a8c440

PR #919, opened 2023-01-19
Engines Connector

joemcelroy:
This is a fork of the Elasticsearch connector with some key differences:
- Overrides the default Query Processor for Searchkit to use a single `combined_fields` query.
- The connector configuration is `apiKey`, `engineName` and `host`. All three are needed. We do not allow to add custom headers.
- Changed the client meta headers to `engine-connector`

Updated the tests.

### DIFF
--- a/packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
+++ b/packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
@@ -102,21 +102,21 @@ export default async function handleRequest(
   ).executeSuggestions(state.searchTerm);
 
   const results: AutocompleteResponseState = response.reduce(
-    (sum, suggestion) => {
+    (acc, suggestion) => {
       const { identifier } = suggestion;
 
       if (identifier === "hits-suggestions") {
         return {
-          ...sum,
+          ...acc,
           autocompletedResults: suggestion.hits.map(fieldResponseMapper)
         };
       } else if (identifier.startsWith("suggestions-completion-")) {
         const name = identifier.replace("suggestions-completion-", "");
 
         return {
-          ...sum,
+          ...acc,
           autocompletedSuggestions: {
-            ...sum.autocompletedSuggestions,
+            ...acc.autocompletedSuggestions,
             [name]: suggestion.suggestions.map((suggestion) => {
               return {
                 suggestion: suggestion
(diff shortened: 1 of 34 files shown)


packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const transporter = new EngineTransporter(host, engineName, apiKey);
+
+  const response = await Searchkit(
+    searchkitConfig,
+    transporter
+  ).executeSuggestions(state.searchTerm);
+
+  const results: AutocompleteResponseState = response.reduce(
+    (sum, suggestion) => {
```
sloanelybutsurely, line 105: `sum` was a bit confusing to me first. I was expecting numbers. Do you mind using `acc` like you've done [elsewhere](https://github.com/elastic/search-ui/pull/919/files#diff-c60b287c9253b1948b77d2a62817e2112b694ba6f9a6b8f4ef897c26525827c7R13)

packages/search-ui-engines-connector/src/handlers/search/Request.ts
```diff
@@ -0,0 +1,58 @@
...
+  size: number;
+  sort: string;
+}
+
+export function getFilters(
+  filters: Filter[] = [],
+  baseFilters: Filter[] = []
+): MixedFilter[] {
+  return filters.reduce((acc, f) => {
+    const isBaseFilter = baseFilters.find((bf) => bf === f);
```
sloanelybutsurely, line 19: nit: i think this could be `baseFilters.includes(f)`

sloanelybutsurely (approved):

packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
joemcelroy: Good suggestion. thanks!

packages/search-ui-engines-connector/src/handlers/search/Request.ts
joemcelroy: agreed. Thanks!

joemcelroy:
@sloanelybutsurely i've also applied the changes to elasticsearch-connector too. Thank you for reviewing!

sloanelybutsurely (approved):

PR #929, opened 2023-02-08
Engines Connector: Allow Custom Transporter 

joemcelroy:
Adding search preview page in kibana, a few issues came up:
- we need to be able to specify custom headers or use kibana's core http service
- we need to change the url to call the internal proxied API
- we cannot specify an api key. Kibana auth (when made through the core http service) will handle authentication concern.

I decided its best to just allow an instance of a transporter to handle the interaction between the API and Search-UI. This way we can use:
- kibana's core http
- specify configuration thats relevant to the transporter

So within kibana, the configuration will be:

```ts
class InternalEngineTransporter implements Transporter {
  constructor(private http: HttpSetup, private engineName: string) {}

  async performRequest(request: SearchReqest) {
    const url = `/internal/enterprise_search/engines/${this.engineName}/search`;

    const response = await this.http.post<SearchResponse>(url, {
      body: JSON.stringify(request),
    });
    return response;
  }
}

  const transporter = new InternalEngineTransporter(http, engineName);
  const connector = new EnginesAPIConnector(transporter);

```

### DIFF
--- a/packages/search-ui-engines-connector/src/handlers/__test__/transporter.test.ts
+++ b/packages/search-ui-engines-connector/src/handlers/__test__/transporter.test.ts
@@ -50,46 +50,4 @@ describe("EngineTransporter", () => {
         expect(response.hits.hits[0]._source.title).toEqual("My title");
       });
   });
-
-  it("perform a request overriding getRoute", () => {
-    const transporter = new EngineTransporter(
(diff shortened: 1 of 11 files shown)


sloanelybutsurely (approved):

joemcelroy:
published as 1.19.1

PR #959, opened 2023-04-06
Update publish_docs_co.yml with new Org Secrets

scottybollinger:
We are refactoring our docs org secrets to segment our public from private docs. This PR updates the keys to the new ones.

After this is completed, @gtback will change the following secrets to private internal repos:

```
VERCEL_GITHUB_TOKEN
VERCEL_TOKEN
VERCEL_ORG_ID
```

### DIFF
--- a/.github/workflows/publish_docs_co.yml
+++ b/.github/workflows/publish_docs_co.yml
@@ -25,7 +25,7 @@ jobs:
       # Docsmobile project dir
       repo: docs.elastic.co
     secrets:
-      VERCEL_GITHUB_TOKEN: ${{ secrets.VERCEL_GITHUB_TOKEN }}
-      VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
-      VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+      VERCEL_GITHUB_TOKEN: ${{ secrets.VERCEL_GITHUB_TOKEN_PUBLIC }}
+      VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN_PUBLIC }}
+      VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID_PUBLIC }}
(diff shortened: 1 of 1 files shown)


scottybollinger:
@gtback not sure if you were pinged or not, but this is the issue to track.

gtback:
I added the label, but it didn't seem to kick off a GitHub action. I think that's because this doesn't actually modify the MDX or other content files, which are the only files that are [triggers](https://github.com/elastic/search-ui/blob/main/.github/workflows/publish_docs_co.yml#L4-L13) for the workflow. Regardless, this would have run using the workflow definition that's already in `main` (without these changes), so it wouldn't have been a good test.

I think we should plan to merge this (once approved), then immediately open a new PR to test the workflow works correctly with the new secrets. After that I'll update the configuration for the org secrets.

gtback (approved):

scottybollinger:
> @scottybollinger I'm fine if you want to wait for one of the @elastic/search-ui-maintainers to approve this as well, but I think it's relatively low-risk to merge as long as we're prepared to test and roll back if there are any problems.

Thanks. I don't have access to merge it. Will wait on one of the maintainers to do so.

sloanelybutsurely (approved):

sloanelybutsurely:
> I added the label, but it didn't seem to kick off a GitHub action. I think that's because this doesn't actually modify the MDX or other content files, which are the only files that are [triggers](https://github.com/elastic/search-ui/blob/main/.github/workflows/publish_docs_co.yml#L4-L13) for the workflow. Regardless, this would have run using the workflow definition that's already in `main` (without these changes), so it wouldn't have been a good test.
> 
> I think we should plan to merge this (once approved), then immediately open a new PR to test the workflow works correctly with the new secrets. After that I'll update the configuration for the org secrets.

@gtback @scottybollinger all set!

gtback:
Thanks, @sloanelybutsurely !

Issue #996, opened 2023-10-05
Elasticsearch connector: Incorrect handling of falsy values


applicity:
### Description of the bug

The elasticsearch connector incorrectly handles non true values in the search results returned.  For example if we have a boolean field called important in our index, when we return a document which has a false value we get an empty object returned.

For example the returned document would look like this:

{ premium: {} }

We would expect this to return:

{ premium: { raw: false } }

Which is the structure of data being returned for a document with a true value, for example we get

{ premium: { raw: true } }

**To Reproduce**

1. Create an index with a boolean field
2. Add a document which has a false value for the boolean field
3. Return the document as a search result
4. The structure of the returned document will have an empty object 

**Expected behavior**

We would expect the field to contain the raw value of the field in all cases, rather than just truthy cases.  With the current behaviour we loose data for fields which have values of false, 0, "" etc.  

**Which backends and packages are you using:**
Backend: [Elasticsearch]
Packages: [react-search-ui, search-ui-elasticsearch-connector]

applicity:
I have created a PR and tests which resolve this issue.

joemcelroy:
thanks for the bug report and PR. Will try release soon.

PR #997, opened 2023-10-05
fix: Issue 996 - falsy values in search results

applicity:
## Description

Update the elasticsearch connector to fix mishandled falsy values

Any field which evaluates to false is returned as an empty object. This change fixes this to return the expected { raw: field value } construct.

## List of changes

Changes the mapping of fields to check for existence rather than truth.

## Associated Github Issues

#996


joemcelroy (approved):

joemcelroy:
thanks! Will try get it released soon.

Issue #1008, opened 2024-01-12
live demo is broken and local run also has issues


swimhiking:
Live demo TypeScript error:
  TypeError
  (0 , $csb__ic_apm_rum_core.observeUserInteractions) is not a function

Local run error:
  path C:\temp\practice\search-ui\packages\react-search-ui
  npm ERR! command failed
  npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c npm run build
  npm ERR! > @elastic/react-search-ui@1.20.2 build
  npm ERR! > npm run clean && npm run build-js
  npm ERR!
  npm ERR!
  npm ERR! > @elastic/react-search-ui@1.20.2 clean
  npm ERR! > rimraf lib

sammy-b-123:
I'm also experiencing the same issues. Did you work out a way around it?

ColeDCrawford:
Same - both the demos at https://docs.elastic.co/search-ui/guides/using-with-vue-js and https://docs.elastic.co/search-ui/overview seem to be broken.

TattdCodeMonkey:
#1010 Should resolve the react issue with the live demo.

The vue issue seems to be different and I will try to look into it separately as well as the typescript issue.

joemcelroy:
Now merged and working. Thanks for the spot! https://codesandbox.io/p/sandbox/github/elastic/search-ui/tree/main/examples/sandbox?file=%2Fsrc%2Fpages%2Felasticsearch%2Findex.js

ColeDCrawford:
Thanks @joemcelroy - is there a separate issue for the problem with the Vue demo?

PR #1010, opened 2024-01-23
[Demo] remove apm init as it is broken

TattdCodeMonkey:
## Description
The `initApm()` call is failing and causing the live demo to not load, removing it for now.

## List of changes
- Remove the `initApm()`

## Associated Github Issues
- #1008


joemcelroy:
I suggest we also remove the deps and instrumented route. See PR https://github.com/elastic/search-ui/pull/764/files

joemcelroy (approved):

Issue #1027, opened 2024-03-14
Upgrade peer-dependencies in react-search-ui to React & React-DOM version 18.2.0


ThunbergOlle:
Hello! 👋  I'm migrating a project to react version 18.2 and saw that we receive peer-dependency warnings when installing about elastic/search-ui only supporting up to react version 18.0.0. Reading the changelogs for version [18.1](https://github.com/facebook/react/blob/main/CHANGELOG.md#1810-april-26-2022) and [18.2](https://github.com/facebook/react/blob/main/CHANGELOG.md#1820-june-14-2022) there does not seem to be any newly introduced changes affecting this repository. 

Is it possible for to update the [peer-dependencies to version 18.2](https://github.com/yakhinvadim/search-ui/blob/3533b13d44e00767fed2717a833998c2afd05313/packages/react-search-ui/package.json#L38-L41)? 😄

joemcelroy:
Yes thanks for letting us know!

joemcelroy:
released in 1.21.1

PR #1028, opened 2024-03-14
Tweak peerDependencies for react 18.2.0

joemcelroy:
- updates the sandbox to use react 18
- update the peerDependencies to react 18

fixes https://github.com/elastic/search-ui/issues/1027


yansavitski (approved):

joemcelroy:
just updated the peerDependencies vs upgrading to react 18. The reason is down to our tests using enzyme and its lack of support for 18. Plan is to use RTL in future.

PR #1034, opened 2024-03-29
Fix pagination styling

### LINKED ISSUE #1032: Pagination styling changed after upgrade
**Describe the bug**
After upgrading to v1.21 from v1.20, we noticed a perhaps unintended change in pagination styling. See images below. 

**To Reproduce**

Pagination styling is different w/ upgrade (e.g. hovering over jump button to next 5 pages or borders around previous and next page buttons). 

**Expected behavior**
Similar look and feel to prior version.

**Screenshots**
Pre-upgrade:
image

After upgrade:
image

Backend: Elasticsearch 
Packages: react-search-ui, react-search-ui-views, search-ui-elasticsearch-connector

miguelgrinberg:
Fixes #1032

The upgrade of rc-pagination caused a few of the custom pagination styles to break. I have updated the styles and now pagination looks as follows:

Screenshot 2024-03-29 at 12 56 56 AM

I'm not sure if this is exactly what was intended, so let me know if any colors need to be different. I'm also not using any hover color changes. If you want that, either for the prev/next buttons or the page numbers, let me know what color or color variable you want.

### DIFF
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
@@ -2,9 +2,11 @@
 @include block("paging") {
   > li {
     border: none;
+    background: transparent;
 
     .rc-pagination-item-link {
       border: none;
+      background: transparent;
       font-size: 16px;
       color: $facetLinkColor;
 
--- a/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
+++ b/packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
@@ -2,14 +2,27 @@
 @include block("paging") {
   > li {
     border: none;
-    background: transparent;
-    outline: none;
+
+    .rc-pagination-item-link {
+      border: none;
+      font-size: 16px;
+      color: $facetLinkColor;
+
+      &:hover {
+        color: $facetLinkColor;
+        background: $pagingHoverBackground;
+      }
+    }
   }
 
   .rc-pagination-disabled {
-    a {
+    .rc-pagination-item-link {
       color: #ccc;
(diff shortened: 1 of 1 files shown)


Samiul-TheSoccerFan (dismissed):
LGTM

Screenshot 2024-04-01 at 9 35 02 AM

packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
```diff
@@ -18,14 +31,6 @@
       color: $linkColor;
```
yansavitski, line 31: New version of library using same color on pagination-item on hover event so it seemed like hover doesn't work at all

packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
```diff
@@ -2,14 +2,27 @@
 @include block("paging") {
   > li {
     border: none;
-    background: transparent;
```
joemcelroy, line 5: Still needs transparent, theres a slight white background thats being applied if removed.
Screenshot 2024-04-02 at 13 20 13

joemcelroy:
Theres still some slight differences between the version. [Previous version](https://codesandbox.io/p/sandbox/heuristic-wilbur-jhcs6f) with subtle white background being applied to the navigation links and larger font

packages/react-search-ui-views/src/styles/themes/reference-ui/components/_paging.scss
miguelgrinberg: Good catch! I have updated the css to use transparent background as before.

joemcelroy:
opened a new PR here. https://github.com/elastic/search-ui/pull/1041

PR #1047, opened 2024-05-23
Update publish_docs_co.yml

scottybollinger:
We have added a new public-only secret and this needs to be updated


joemcelroy (approved):

scottybollinger:
Thanks for the review @joemcelroy! Would you mind merging this? I do not have access to do so.

Issue #1046, opened 2024-05-13
Latest versions fail


avoinea:
```
[91mModule not found: Error: Package path ./lib/cjs/helpers is not exported from package /app/node_modules/@elastic/search-ui (see exports field in /app/node_modules/@elastic/search-ui/package.json)
ModuleNotFoundError: Module not found: Error: Package path ./lib/cjs/helpers is not exported from package /app/node_modules/@elastic/search-ui (see exports field in /app/node_modules/@elastic/search-ui/package.json)
    at /app/node_modules/webpack/lib/Compilation.js:2016:28
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:798:13
    at eval (eval at create (/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:10:1)
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:270:22
    at eval (eval at create (/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:434:22
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:116:11
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:670:25
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:855:8
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:975:5
    at /app/node_modules/neo-async/async.js:6883:13
    at /app/node_modules/webpack/lib/NormalModuleFactory.js:958:45
    at finishWithoutResolve (/app/node_modules/enhanced-resolve/lib/Resolver.js:567:11)
    at /app/node_modules/enhanced-resolve/lib/Resolver.js:656:15
    at /app/node_modules/enhanced-resolve/lib/Resolver.js:714:5
    at eval (eval at create (/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:16:1)
    at /app/node_modules/enhanced-resolve/lib/Resolver.js:714:5
    at eval (eval at create (/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:27:1)
    at /app/node_modules/enhanced-resolve/lib/DescriptionFilePlugin.js:89:43
    at /app/node_modules/enhanced-resolve/lib/Resolver.js:714:5
[...]

PR #1050, opened 2024-05-30
Module support updates.

TattdCodeMonkey:
## Description
Updates to `package.json` `exports` to include types and package.json
Additionally updated to TS v4.9 as during investigation we found that module support was not fully available in TS until v4.7

## List of changes
- Added `"types": "./lib/esm/index.d.ts"` to `"exports"` configs
- Updates TS to v4.9.3
- Misc. type and dep fixes

## Associated Github Issues
#1046


joemcelroy (dismissed):
can confirm that helpers is accessible from `import { helpers } from "@elastic/search-ui"` through my test typescript vite app and using npm link to link search-ui package.

joemcelroy (approved):

PR #1063, opened 2024-06-24
fix: issue 1062 - filters applied more than once

### LINKED ISSUE #1062: [elasticsearch-connector] base filters are incorrectly applied twice
**Describe the bug**
When defining base filters as part of the search configuration they are applied twice into the query.  If the base filter is a "none" query the second time it is applied then it will be applied as a "must" and therefore you get zero results.

**To Reproduce**

Define a search query with a filter like this: 

```
...
filters: [{ field: 'uri', values: ['one'], type: 'none' }]
```

The elastic search query is built as:

```
{
  "bool": {
    "filter": [
      {
        "bool": {
          "must_not": [
            {
              "term": {
                "subject_uri": "one"
              }
            }
          ]
        }
      },
      {
        "bool": {
          "filter": [
            {
              "term": {
                "subject_uri": "one"
              }
            }
          ]
        }
      }
    ],
    "must": []
  }
}

``` 

**Expected behavior**

I would expect the query only to have the filter added once, and in the case of the "none" version that this be the must_not version.

Backend: [Elasticsearch]
Packages: [react-search-elasticsearch-connector]

applicity:
The test uses the same array twice so the check for filters.includes() works.  In the real world the baseFilter and the filter are different.  So I have updated the test and changed the check to use basefilters.find().

Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

## Description
This PR addresses the problem with filters being applied more than once.  There is a test for this but it is broken as it uses the same value in two arrays to check for inclusion.  I have updated the test and the code in getFilters.

## List of changes

Updated the test to fix the assumption of two values.
Changed the code in getFilters to use a find rather than includes which will check for the filter.field equality

## Associated Github Issues

fixes #1062

### DIFF
--- a/packages/search-ui-elasticsearch-connector/src/handlers/search/Request.ts
+++ b/packages/search-ui-elasticsearch-connector/src/handlers/search/Request.ts
@@ -16,7 +16,7 @@ export function getFilters(
   baseFilters: Filter[] = []
 ): MixedFilter[] {
   return filters.reduce((acc, f) => {
-    const isBaseFilter = baseFilters.includes(f);
+    const isBaseFilter = baseFilters.find((base) => base.field === f.field);
     if (isBaseFilter) return acc;
 
     const subFilters = f.values.map((v) => {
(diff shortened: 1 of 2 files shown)


applicity:
Has the cla-checker-service lost track of signed agreements.

joemcelroy (approved):

Issue #1109, opened 2025-02-07
[react-search-ui] setSort not working, Sorting not working


craquet:
**Describe the bug**
The SearchContext of react-search-ui provides a method called `setSort` to sort the results. It provides a signature for setting the sorting like this:

```ts
search.driver.setSort("field", "asc") // Version 1
```

and like this:

```ts
search.driver.setSort([{ field: "field", direction: "asc" }], "asc") // Version 2
```

There are two problems here

1. When specifying the sort using Version 1, it does not work (does nothing, as described in #932 , which is closed but not fixed)
2. When specifying the sort using Version 2, the second parameter (`"asc"`) is obsolete and should not have to be specified. Though I assume it has no effect anyways.

**To Reproduce**

Expose the search context to the window object and try the above methods using the browser console. Using Version 1 will update the state, but will not issue a new search. Version 2 works, although the method signature is wrong.

Also See #932 as this issue also applies here (the `Sorting` component does not work).

**Expected behavior**
Version 1 and Version 2 as outlined above should work as expected and issue a new search with the correct sorting.

The `Sorting` component should behave as expected and issue a new search when a sort option is selected.

**Related**
See also #932 

**Which backends and packages are you using:**
Backend: [Elasticsearch]
Packages: [react-search-ui]

sphilipse:
Hi @craquet! Thanks for reporting this bug, we'll prioritize this and try to fix it. I can't promise a definite date by which this will be fixed, but it's definitely on our radar.

PR #1112, opened 2025-02-21
[ElasticSearch] Fix sorting

yansavitski:
## Description
Fix sorting when has sortField and sortDirection values in state

## Associated Github Issues
https://github.com/elastic/search-ui/issues/1109


TattdCodeMonkey (approved):
If I understood how sorting in SearchKit works then these changes LGTM. But I did not get a chance to test this myself.

PR #1114, opened 2025-02-28
Fix export modules

yansavitski:
## Description
The `exports` field in `package.json` forces ESM modules to use ESM files and CommonJS modules to use CJS files. However, **Node.js requires file extensions in ESM imports** ([see Node.js docs](https://nodejs.org/api/esm.html#mandatory-file-extensions)), which caused **backwards compatibility issues** for users who previously imported `search-ui` in Node.js as a CommonJS module.

To solve this and **ensure compatibility**, we added **bundling** to generate a single file per module. This approach also provides additional benefits, such as improved module resolution and reduced file system lookups.

## List of changes
- Add **tsup** for bundling
- Use common **tsup** configuration
- Use common **TS** configuration
- Update export **lib** structure
- Change import searchkit/sdk as ESM and CJS always use commonJS version for this module
- Improved **watch** mode for all packages

## Associated Github Issues
https://github.com/elastic/search-ui/issues/1046


packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
```diff
@@ -4,15 +4,20 @@ import type {
...
+import type { SearchkitConfig } from "@searchkit/sdk";
+
+const {
+  default: Searchkit,
+  CompletionSuggester,
+  HitsSuggestor,
+  PrefixQuery
+} = typeof SearchkitModule.default === "object"
+  ? (SearchkitModule.default as unknown as typeof SearchkitModule)
+  : (SearchkitModule as typeof SearchkitModule);
```
TattdCodeMonkey, line 20: we're having to do this bit of code several times, would it make sense to wrap this bit of logic either in a file or write a utility function for it?

Either

```typescript
import * as SearchkitSDK from "@searchkit/sdk";

export const SearchkitModule = typeof SearchkitSDK.default === "object"
  ? (SearchkitSDK.default as unknown as typeof SearchkitSDK)
  : (SearchkitSDK as typeof SearchkitSDK);
```

Or a utility function.
```typescript
import * as SearchkitModule from "@searchkit/sdk";

export function handleSearchImport(module: SearchkitModule): SearchkitModule {
  return typeof SearchkitModule.default === "object"
  ? (SearchkitModule.default as unknown as typeof SearchkitModule)
  : (SearchkitModule as typeof SearchkitModule);
}
```

It'd be nice if we didn't have to do all the casting, I bet there is a way to describe this in a override `.d.ts` file but I'd have to some research to figure that out.

TattdCodeMonkey (approved):
One comment about the search kit import, but I think we want to remove that dependency so if we don't want to spend extra time handling it better I understand.

packages/search-ui-engines-connector/src/handlers/autocomplete/index.ts
yansavitski: I suppose d.ts would ignored in that case, as we need this manipulations mostly for import in esm modules and require for cjs and d.ts just ignored in that case because it's missed in .js bundle after all

TattdCodeMonkey (approved):

PR #1138, opened 2025-04-03
Add prod example to sandbox

yansavitski:
## Description
Add ProductionReady Example to a sandbox.
Sandbox Link - https://codesandbox.io/p/sandbox/github/elastic/search-ui/tree/add-prod-example-to-sandbox/examples/sandbox

## List of changes
- Add new recommended example to home page
- Run nodejs API server and vite on the same task
- Calculate sandbox api in front
- Fix UI issues

Before:
image
image

Actual:
image


mdefazio:
Tiny nit, I wonder if we can think of a better name than Acme, inc if we're showing national park examples.  
Hiking Inc. ? Parks Inc? Not sure, but that just stood out to me 😅

mdefazio:
Thank you!

examples/sandbox/.codesandbox/tasks.json
```diff
@@ -7,27 +7,19 @@
     },
     {
       "name": "Build App",
-      "command": "yarn build"
+      "command": "VITE_IS_SANDBOX=true yarn build"
     }
   ],
 
   // These tasks can be run from CodeSandbox. Running one will open a log in the app.
```
TattdCodeMonkey, line 14: `.json`, doesn't support comments. This would need to be a `.jsonc` file if we want to add comments. does codesandbox support `.jsonc`?

examples/sandbox/src/pages/ecommerce/styles.scss
```diff
@@ -0,0 +1,31 @@
+@tailwind base;
```
TattdCodeMonkey, line 1: why do we need to change to sass from vanilla css here?

examples/sandbox/.codesandbox/tasks.json
yansavitski: Don't think so
https://codesandbox.io/docs/learn/devboxes/task

I'll remove comments then

examples/sandbox/src/pages/ecommerce/styles.scss
yansavitski: Previously `styles.css` applied styles globally to the search-ui classes and affected other examples. So I added a wrapper `ecommerce-container` to apply styles only for this example. To avoid repeating that logic in plain css, I rewrote it using scss.

TattdCodeMonkey (approved):

PR #1143, opened 2025-04-29
Replace searchkit package

yansavitski:
## Description
This PR refactors the Elasticsearch connector to use a new ApiClient architecture, removing the SearchKit/SDK dependency. The changes improve the connector's maintainability and make it easier to integrate with non-JavaScript backends and reduced the bundle size. The new architecture allows for better customization of requests and responses while maintaining compatibility with Search UI's core functionality.

**Before (with Searchkit):**
Search UI → State & Config → ElasticsearchConnector → (transform to SearchKit) → SearchKit → Elasticsearch Request → SearchKit → ElasticsearchConnector → (transform to Search UI) → Search UI

**After (without SearchKit):**
Search UI → State & Config → ElasticsearchConnector → Elasticsearch Request → ElasticsearchConnector → Search UI

## List of changes
- Removed SearchKit/SDK dependency
- Added new ApiClient architecture for better flexibility
- Added QueryBuilder for constructing Elasticsearch queries
- Added response and filter transformers
- Updated handlers and autocomplete generation
- Added comprehensive test coverage
- Updated documentation with new ApiClient usage

**Sandbox link**: https://codesandbox.io/p/sandbox/github/elastic/search-ui/tree/replace-searchkit-package/examples/sandbox


packages/search-ui-elasticsearch-connector/src/transporter/ApiClientTransporter.ts
```diff
@@ -0,0 +1,53 @@
...
+      throw new Error("Host or cloud is required");
+    }
+
+    let host = this.config.host;
+
+    if (this.config.cloud) {
+      host = getHostFromCloud(this.config.cloud);
+    }
+
+    const response = await fetch(host + "/" + this.config.index + "/_search", {
```
TattdCodeMonkey, line 38: (non-blocking)
it could be worth using `URL` for the validation
```suggestion
    const response = await fetch(new URL(`${this.config.index}/_search`, host), {
```

Or if theres any performance concerns there a template string here would read a little cleaner IMO

```suggestion
    const response = await fetch(`${host}/${this.config.index}/_search`, {
```

packages/search-ui-elasticsearch-connector/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    ([fieldKey, fieldConfiguration]) =>
+      `${fieldKey}^${fieldConfiguration.weight || 1}`
+  );
+
+export const getHostFromCloud = (cloud: { id: string }): string => {
+  const { id } = cloud;
+  // the cloud id is `cluster-name:base64encodedurl`
+  // the url is a string divided by two '$', the first is the cloud url
+  // the second the elasticsearch instance, the third the kibana instance
+  const cloudUrls = atob(id.split(":")[1]).split("$");
```
TattdCodeMonkey, line 36: should we update this `atob` usage to `Buffer.from(data, 'base64')` since atob is deprecated?

TattdCodeMonkey: Also do we have any validation on the cloud id to ensure this wont error at runtime if they user configs a bad cloud id?

packages/search-ui-elasticsearch-connector/src/transporter/__tests__/ApiClientTransporter.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  });
+
+  it("should use cloud host when cloud config is provided", async () => {
+    const cloudConfig = {
+      id: "test-cloud-id",
+      username: "test-user",
+      password: "test-password"
+    };
+
+    (getHostFromCloud as jest.Mock).mockReturnValue("https://cloud-host.com");
```
TattdCodeMonkey, line 96: we don't have any tests for `getHostFromCloud` so it could be good to not mock this an test with a stubbed cloud id here. and also write some unit tests for invalid cloud ids to make sure it errors in ways we expect. IE gives useful error messages if the cloud id fails to decode.

TattdCodeMonkey:
The only updates I want from my comments are getting rid of the type casting. otherwise this looks good from what I can understand. I did not test the code though.

docs/reference/api-connectors-elasticsearch.md
```diff
@@ -62,13 +76,14 @@ const connector = new ElasticsearchAPIConnector(
...
+| Param             | Type   | Description                                                                                                                                                                                                                                                                                                                                            |
+| ----------------- | ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| cloud             | object | **Required if `host` not provided.** Object type. The cloud id for the deployment within elastic cloud.                                                                                                                                                                                                                                                |
```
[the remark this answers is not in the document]
yansavitski: Yeah agree that we should better show expected types in doc

packages/search-ui-elasticsearch-connector/src/handlers/handleAutocomplete.ts
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  const buildResultConfig = (
+    config: QueryConfig | ResultSuggestionConfiguration,
+    type?: string
+  ): SuggestionConfig => {
+    const builder = new ResultsAutocompleteBuilder(
+      state,
+      config,
+      (config as QueryConfig).resultsPerPage ||
```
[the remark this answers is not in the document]
yansavitski: Yeah we don't need it, I probably first added this changes and then extended `ResultSuggestionConfiguration` with `resultsPerPage`

packages/search-ui-elasticsearch-connector/src/transporter/ApiClientTransporter.ts
yansavitski: Applied version with URL

packages/search-ui-elasticsearch-connector/src/utils.ts
yansavitski: added tests and two implementations atob for browser and Buffer for node

packages/search-ui-elasticsearch-connector/src/transporter/__tests__/ApiClientTransporter.test.ts
yansavitski: Added utils tests for node and browser env

TattdCodeMonkey (approved):

PR #1146, opened 2025-04-30
Add separate entry point for ApiProxyConnector

yansavitski:
## Description
Added a separate entry point for the ApiProxyConnector to allow users to import only the necessary code, reducing the bundle size in production environments.

## List of changes
- Added new entry point `api-proxy.ts` for ApiProxyConnector
- Updated tsup config to support multiple entry points
- Added exports configuration in package.json
- Updated documentation with import methods and bundle size optimization tips


docs/reference/api-connectors-elasticsearch.md
```diff
@@ -92,6 +92,8 @@ The `ApiProxyConnector` is used when you want to proxy search requests through y
 It sends `onSearch` and `onAutocomplete` requests to your API, which is expected to forward them to Elasticsearch using `ElasticsearchAPIConnector`.
 
 ```js
+import { ApiProxyConnector } from "@elastic/search-ui-elasticsearch-connector/api-proxy";
+// OR
 import { ApiProxyConnector } from "@elastic/search-ui-elasticsearch-connector";
```
saarikabhasi, line 97: nit: should we combine this line with OR comment? 

```
// OR 
// import { ApiProxyConnector } from "@elastic/search-ui-elasticsearch-connector";
```

saarikabhasi (approved):
Code only review LGTM

docs/reference/api-connectors-elasticsearch.md
yansavitski: Yeah, I agree, hadn't strong opinion about that, but seems like it makes more sense

saarikabhasi (approved):

PR #1165, opened 2025-05-31
Reduce dependencies from qs

benmccann:
Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

## Description

https://npmgraph.js.org/?q=qs - 18 dependencies
https://npmgraph.js.org/?q=neoqs - 0 dependencies

Used the legacy entrypoint for CJS support

## List of changes

Swapped `qs` for `neoqs`

### DIFF
--- a/packages/search-ui/package.json
+++ b/packages/search-ui/package.json
@@ -32,14 +32,13 @@
   },
   "devDependencies": {
     "@types/history": "^4.7.11",
-    "@types/qs": "^6.9.7",
     "typescript": "^5.8.3"
   },
   "dependencies": {
     "date-fns": "^1.30.1",
     "deep-equal": "^1.0.1",
     "history": "^4.9.0",
-    "qs": "^6.9.7"
+    "neoqs": "^6.13.0"
   },
   "exports": {
     ".": {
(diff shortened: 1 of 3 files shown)


benmccann:
Not stale

benmccann:
not stale

yansavitski:
Hey
Thank you for creating PR, but let's to stick with qs here.

neoqs has around ~14k downloads per week, which is quite low. It doesn’t look widely adopted or strongly maintained by the community. That increases the risk of running into issues later with bugs, security, or lack of support.

qs, on the other hand, is a well-established and widely used library. Even if it means one extra dependency, that’s a much safer trade-off for us than relying on a relatively unknown package.

benmccann:
> Even if it means one extra dependency

Just to be clear, it's 18 extra dependencies

There are a number of other alternatives that may be worth consideration if you find neoqs not widely used enough: https://e18e.dev/docs/replacements/qs

yansavitski:
Thanks for the clarification.

But still, it doesn't look like a strong reason to switch. qs is widely used and proven, while neoqs (or similar options) don’t really have that level of adoption or track record.

Reducing dependencies is nice, but here it feels like we’re replacing something stable with something less known without a clear benefit.

PR #1167, opened 2025-06-04
fix: rawResponse hard-coded to null

rlueder:
## Description

While trying to access `rawResponse.took` I noticed `rawResponse` is always `null` in `@elastic/search-ui-elasticsearch-connector`

## List of changes

- updates `rawResponse` to `response` instead of `null`
- updates necessary tests


yansavitski (approved):
LGTM

PR #1168, opened 2025-06-06
Enhance facet configuration and sorting options

ax-ml:
- Updated `FacetConfiguration` to include `field` and changed `sort` to use `FacetSortOption`.
- Modified `SearchQueryBuilder` to utilize the new `field` property for facet filtering.
- Adjusted tests to reflect changes in facet configuration and ensure compatibility with the new structure.
- Improved `transformFacetToAggs` to handle the updated facet sorting logic.

Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

### DIFF
--- a/packages/search-ui/src/types/index.ts
+++ b/packages/search-ui/src/types/index.ts
@@ -1,5 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
 import type { SearchDriverActions } from "..";
 
 export type SortOption = {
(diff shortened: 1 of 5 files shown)


packages/search-ui/src/types/index.ts
```diff
@@ -153,7 +151,8 @@ export type FacetConfiguration = {
   ranges?: FilterValueRange[];
   center?: string;
   unit?: string;
-  sort?: "count" | "value";
+  sort?: FacetSortOption;
+  field?: string;
```
yansavitski, line 155: What’s the benefit of introducing the new field property here?

At the moment, it seems to introduce additional complexity without clear advantages. We already use the object key (key) to represent the field, so adding a separate field property could create confusion:
	•	It duplicates meaning — both key and field now refer to the same thing.
	•	It’s unclear which one takes precedence if they differ.
	•	There’s potential for naming collisions — e.g. facets: { title: { field: "title" } } is fine, but what happens if the key is different from the field?

Issue #1178, opened 2025-07-07
Async `getQueryFn`


italodirenzo93:
**Is your feature request related to a problem? Please describe.**
My specific use-case is in doing a kNN search with `dense_vector`. I'm using `getQueryFn` to execute a kNN search like so:
```ts
getQueryFn: (state) => {
  return {
    knn: {
      field: 'embedding',
      query_vector_builder: {
        text_embedding: {
          model_id: 'openai-embeddings',
          model_text: state.searchTerm
        }
      }
    }
  }
}
```
This works by using Elasticsearch's [Inference API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-openai). However usage of this API requires a paid Enterprise license.

**Describe the solution you'd like**
Another way to create an input vector for a kNN search is to generate it in your own application code and pass it to to the `query_vector` field. However, since this almost always will be an `async` function call, it cannot be done within `getQueryFn` currently. Ideally, the example code above should be able to be written like this:
```ts
async getQueryFn: (state) => {
  const embedding = await openai.embeddings.create({
    model: 'text-embedding-3-small',
    input: state.searchTerm,
  })

  return {
    knn: {
      field: 'embedding',
      query_vector_builder: embedding.data[0].embedding
    }
  }
}
```

**Describe alternatives you've considered**
The `interceptSearchRequest` hook is asynchronous. Theoretically, one could create the base query within `getQueryFn` and then set the `query_vector` from within `interceptSearchRequest`. However, it feels more appropriate to keep code related to the base query within `getQueryFn` so that `interceptSearchRequest` can be kept for filters and other edge cases.

PR #1181, opened 2025-07-17
Allow getQueryFn to return a promise.

italodirenzo93:
## Description
Converts `getQueryFn` to an `async` function so that asynchronous work can be done when creating the query.

## Associated Github Issues
#1178


yansavitski (approved):
LGTM

Thank you

Issue #1190, opened 2025-09-11
ElasticsearchAPIConnector searchTerm with white space throws error


lluisemper:
**Describe the bug**
If to the searchTerm I add a whitespace the package search-ui-elasticsearch-connector throws and error internally.

**To Reproduce**
By doing a post request using postman to `"/api/search"` it will throw an error if the searchTerm includes a white space between 2 characters. If I remove the white space it works as it should.

I created a small express server to reproduce this:
```js
import express from "express";
import cors from "cors"; // Import the cors package
import ElasticsearchAPIConnector from "@elastic/search-ui-elasticsearch-connector";

const app = express();
const PORT = process.env.PORT || 3001;

app.use(
  cors({
    origin: "*",
  })
);
app.use(express.json());
app.use(express.urlencoded({ extended: true }));

app.post("/api/search", async (req, res) => {
  try {
    const queryConfig = {
      result_fields: { log: { raw: {} } },
      search_fields: { log: {} },
      facets: {},
    };
    const state = {
      current: 1,
      filters: [],
      resultsPerPage: 20,
      searchTerm: "T 0", // WIHTOUT WHITESPACES IT WORKS WELL
      sortDirection: "",
      sortField: "",
    }; // Use req.body to access the parsed JSON

    const connector = new ElasticsearchAPIConnector({
      host: "<host>",
      index: `<index>`,
      apiKey: "<apiKey>",
    });
    const response = await connector.onSearch(state, queryConfig);
    return res.json(response);
  } catch (error) {
    console.error("🚀 ~ error:", error);
    return res.status(500).json({ error: "Something went wrong" });
  }
});

// Start the server
app.listen(PORT, () => {
  console.log(`Server is running on http://localhost:${PORT}`);
});

```

This is the error I get:
```
🚀 ~ error: TypeError: Cannot read properties of undefined (reading 'total')
    at de (file://test/node_modules/@elastic/search-ui-elasticsearch-connector/lib/index.mjs:1:6891)
    at J (file://test/node_modules/@elastic/search-ui-elasticsearch-connector/lib/index.mjs:1:6396)
    at file://test/node_modules/@elastic/search-ui-elasticsearch-connector/lib/index.mjs:1:12147
    at Generator.next (<anonymous>)
    at o (file://test/node_modules/@elastic/search-ui-elasticsearch-connector/lib/index.mjs:1:493)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

**Expected behavior**
[...]

PR #1191, opened 2025-09-11
#1190 handle errors from response in search-ui-elasticsearch-connector

lluisemper:
Before submitting this PR:

   to this repository.
   applicable).
   storybook (if applicable).

## Description
Handle api error in ElasticsearchAPIConnector
## List of changes
If an error is found, throw it instead of failing silently.
## Associated Github Issues
https://github.com/elastic/search-ui/issues/1190

### DIFF
--- a/packages/search-ui-elasticsearch-connector/src/handlers/handleSearch.ts
+++ b/packages/search-ui-elasticsearch-connector/src/handlers/handleSearch.ts
@@ -7,10 +7,16 @@ import { SearchQueryBuilder } from "../queryBuilders/SearchQueryBuilder";
 import { transformSearchResponse } from "../transformer/responseTransformer";
 import type { IApiClientTransporter } from "../transporter/ApiClientTransporter";
 import type {
+  ElasticsearchError,
   PostProcessRequestBodyFn,
   RequestModifiers,
-  SearchRequest
+  SearchRequest,
+  SearchResponseWithError
 } from "../types";
+
+interface ElasticsearchSearchError extends Error {
+  elasticsearchError: ElasticsearchError;
+}
 
 export const handleSearch = async (
   state: RequestState,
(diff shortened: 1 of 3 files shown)


packages/search-ui-elasticsearch-connector/src/handlers/handleSearch.ts
```diff
@@ -37,6 +43,17 @@ export const handleSearch = async (
       (requestBody: SearchRequest) => apiClient.performRequest(requestBody)
     );
   }
+  if (response.error) {
```
TattdCodeMonkey: I think this would be better handled using `interceptSearchRequest` so that you can throw errors specific to your needs instead of being handled here in the library

packages/search-ui-elasticsearch-connector/src/handlers/handleSearch.ts
```diff
@@ -37,6 +43,17 @@ export const handleSearch = async (
...
+    const rootCause = response.error.root_cause?.[0];
+    const message =
+      rootCause?.reason ||
+      response.error.reason ||
+      "Elasticsearch search failed";
+
+    const error = new Error(message) as ElasticsearchSearchError;
```
TattdCodeMonkey: instead of casting I would create a class for `ElasticsearchSearchError` and pass the `response.error` to it.

TattdCodeMonkey (commented):
The logic you're adding is very specific and I don't think it belongs here.

packages/search-ui-elasticsearch-connector/src/handlers/handleSearch.ts
lluisemper: As with the error above, it is not clear that the library breaks because the request contains an error. And you end up trying a few different things untill you find out the root of the issue

